PR #3, opened 2016-09-27
Clean up the package.json file

blgm:
There were some extra things in the `package.json` file which look like they are no longer needed.


mattbaileyuk:
@blgm Good spot on things to remove; feels good to have a lighter `package.json` where possible

PR #8, opened 2016-09-29
Add an npm download numbers badge

blgm:
This displays the number of downloads and also is a handy link to the `npm` page from GitHub.


mattbaileyuk:
Thanks for the pull request. Having discussed this with others, we think that while the link to `npm.js` could be useful, we don't think a download numbers badge is necessary, at least in the early days of the project. Maybe something to come back to in future.

PR #9, opened 2016-09-29
Add a dev dependencies badge as well

blgm:
See PR #7 for previous addition of badge for non-dev dependencies. It's also worth showing that dev dependencies are kept up to date.


andrew-coleman:
Many thanks for the pull request.  After much discussion, it was decided that the inclusion of a dependencies badge might confuse users who only use JSONata within a browser.

PR #11, opened 2016-10-05
Added sum, min, max, count, average functions, tests and their proper documentation

vladbalanescu:


andrew-coleman:
Hi Vlad, thanks for the pull request.  A few things:
  - There are many changes to the formatting/spacing of code that is not affected by your change.  Please don't do that.
  - For all functions, passing in undefined should return undefined, and not throw an error. This is so that expressions that take a path expression don't blow up if the path doesn't match anything (this is a query language).
  - The docs for the functions read like they are casting functions.  We'll need to work on that.

PR #13, opened 2016-10-05
added functions $length, $split, $join

andrew-coleman:


jsonata.js
```diff
@@ -2013,6 +2013,137 @@ function functionUppercase(str) {
...
+        };
+    }
+
+    return str.split(separator, limit);
+}
+
+/**
+ * Join an array of strings
+ * @param {Array} strs - array of string
+ * @param {String} separator - the token that splits the string (optional)
```
mattbaileyuk: You can use `[separator]` to indicate to JSDoc that the `separator` parameter is optional

jsonata.js
```diff
@@ -2013,6 +2013,137 @@ function functionUppercase(str) {
...
+    }
+
+    return str.length;
+}
+
+/**
+ * Split a string into an array of substrings
+ * @param {String} str - string
+ * @param {String} separator - the token that splits the string
+ * @param {Integer} limit - max number of substrings (optional)
```
mattbaileyuk: You can use `[limit]` to indicate to JSDoc that the `limit` parameter is optional

PR #15, opened 2016-10-07
cast functions: undefined arg returns undefined

andrew-coleman:
Added to the casting functions the usual condition that invoking them on undefined (no match) should return undefined


mattbaileyuk:
Reviewed and merging. This should complete the next block of functions, so might be time to up the version so these get into npm.

PR #16, opened 2016-10-11
Node v010

andrew-coleman:


mattbaileyuk:
Discussed use of multiline strings with @andrew-coleman as we don't have the backtick multiline support of ES6 in node 0.10, but think what's there now looks just as neat.

PR #17, opened 2016-10-12
Remove dependency badge

mattbaileyuk:
The `david-dm` dependency badge seems to not be working more than it is; I think their servers are struggling, but in turn it just gives an ugly `image not found` on the front page of this repo.

As there are no dependencies, the badge is only indicating as such, and is not actually making anyone aware of out of date dependencies (as you would see when it's used elsewhere). I think it should be removed for now, and could return in the future if any dependencies are added.


mattbaileyuk:
Merging after discussing with @andrew-coleman

PR #23, opened 2016-11-03
New functions and ability to select multiple indexes in a predicate

andrew-coleman:


mattbaileyuk:
Looks good with more functions popping up :smile: While the unit tests provide some good examples of functions, I think we could do with soon adding examples either into `functions.md` or some doc elsewhere.

PR #24, opened 2016-11-07
added "in" operator; fixed := precedence; fixed object constructor be…

andrew-coleman:
…haviour with empty context

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -32,12 +32,13 @@ var operators = {
     '`': 80,
     '**': 60,
     '..': 20,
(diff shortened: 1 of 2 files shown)


mattbaileyuk:
@andrew-coleman It would be good to include an example of `x in [x,y]` in `tutorial.md` at the same time as there are examples of `and` and `or` there

mattbaileyuk:
Discussed with @andrew-coleman and we're going to reorganise and expand `tutorial.md` so will cover this addition in that.

PR #25, opened 2016-11-10
Wrap jsonata in closure to prevent leakage of internal variables

knolleary:
This allows `jsonata.js` to be loaded a browser without exposing its internal functions/variables.

A side-effect is the generated jsdoc no longer includes those internal functions etc. I've found lots of questions on StackOverflow as to how to get jsdoc to pickup such functions, but no real solution.

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -5,2785 +5,2801 @@
  */
 
 'use strict';
+/**
+ * @module JSONata
+ * @description JSON query and transformation language
+ */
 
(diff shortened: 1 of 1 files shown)


knolleary:
Ah - missed the DCO commit message. Let me fix that up.

knolleary:
All fixed - ready for review

andrew-coleman:
Looks good.  Thanks!

PR #29, opened 2016-11-16
Use function version of 'use strict'

blgm:
The `'use strict'` directive is only recognised as the beginning of a script or function.  If JSONata is merged with other files using `browserify` or similar technology, the global `'use strict'` at the top of the file will cease to apply as it will no longer be the first statement in the file.  Therefore it is safer to put `'use strict'` at the start of the JSONata main function, so that it is guaranteed to apply to that function and other functions defined within it.  This is not necessary for the test code, which retains a global `'use strict'`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -40,7 +40,7 @@
       }
     }],
     "semi": ["error", "always"],
-    "strict": ["error", "global"],
+    "strict": ["error", "function"],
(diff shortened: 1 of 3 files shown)


mattbaileyuk:
Seems like an uncontroversial change to me, and a good catch

PR #28, opened 2016-11-15
Consistent quotation style and no eval()

blgm:
Use single quotes except when to use them would require escaping of quotes.
Disallow `eval()` as this is a potential security risk.

PR #30, opened 2016-11-16
ESLint updates, principally to ban eval()

blgm:
Also:
- banned used of `console`, e.g. `console.log()`
- currently the project follows the `1tbs` brace style, so added a rule to stop anyone doing anything different in the future

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,12 +17,15 @@
   // Additional rules
   "rules": {
     "array-bracket-spacing": ["error", "never"],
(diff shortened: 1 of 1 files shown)


mattbaileyuk:
Agree it's good to add this to ESLint rules, even though @andrew-coleman and I would reject PRs with `eval`s just so that it should get picked up by local `npm t` runs and/or Travis before we'd even have to look at the code.

@blgm Merging this, it sounds like you're going to drop https://github.com/jsonata-js/jsonata/pull/28

PR #31, opened 2016-11-17
Changes to behaviour of object constructor and result arrays

andrew-coleman:
User feedback suggests that the mechanism for creating 'grouped' objects versus arrays of objects is confusing.  Currently creating a grouped or merged object is done using the following syntax:
`location.path.{key: value}`
and the syntax for creating arrays of objects is:
`location.path.[{key: value}]`

The confusion arises because the `.` operator is really a 'map' higher order function, so the expectation is that each value in the array on the LHS of `.` should produce a value in a result array without the need for any array syntax.  The function that processes this `.` operator has special case handling for .{} to enable this grouping, and that is probably a Bad Thing. 

So I have removed this 'special case' code and now `location.path.{key: value}` behaves like `location.path.[{key: value}]` used to.  The group-by (merge) capability can now be achieved using the following:
`location.path{key: value}`
There is no `.` operator before the object constructor.

Other feedback related to how the results of expressions are getting returned.  Currently, if multiple values are matched by an expression, they are returned in a flattened array.  If a single value is matched, it is returned directly (not in an array).  This can be nuisance for the host language to process if it is not known whether an expression is going to match a single or multiple values.  The host code would have to query the type first before iterating over the result (or not).  Worse still is if the offending expression is embedded in, say, an object constructor expression such as:
`Phone{type:number}`
In this case we could end up with a result object like this:
```
{
  "home": "0203 544 1234",
  "office": [
    "01962 001234",
    "01962 001235"
  ],
  "mobile": "077 7700 1234"
}
```
 whereby some of the object values are arrays and some are not.

To fix this, I have introduced some new syntax `[]` which, if appended to a step, will force the output to always be an array.  This can be mixed with predicates without any problem. E.g.:

`Phone{type:number[]}`
produces
```
{
  "home": [
    "0203 544 1234"
  ],
  "office": [
    "01962 001234",
    "01962 001235"
  ],
  "mobile": [
    "077 7700 1234"
  ]
}
```

`Phone[type='home'][].number`
produces
`[ "0203 544 1234" ]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,33 +6,52 @@
 JSON query and transformation language
 
 ##Introduction
-The primary purpose of this language is to extract values from JSON documents, with the
-additional capabilities to combine these values using a set of basic functions
-and operators, and also the ability to format the output into any arbitrary JSON structure.
+JSONata is a lightweight query and transformation language for JSON data.
+Inspired by the 'location path' semantics of XPath 3.1, it allows sophisticated
+queries to be expressed in a compact and intuitive notation.  A rich complement of built in
+operators and functions is provided for manipulating and combining extracted
+data, and the results of queries can be formatted into any JSON output structure
+using familiar JSON object and array syntax.
+Coupled with the facility to create user defined functions, advanced expressions
+can be built to tackle any JSON query and transformation task.
+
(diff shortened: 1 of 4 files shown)


mattbaileyuk:
Had previously discussed the implications of these changes, so having reviewed the code we're good to merge (essentially good to make these potentially breaking changes earlier before adoption picks up). Can't rebase as there's conflicts, so will squash and merge instead.

PR #36, opened 2017-01-30
Simplify tokenization of two-char operators

blgm:
This change updates the tokenization of two character operators to be the same as for single character operators.  It reduces the lines of code required, aiding readability.  It uses `hasOwnProperty()` rather than multiple `if()`s, as is the pattern for single character operators.  This may have slightly different performance characteristics, but is unlikely to make a significant difference as `hasOwnProperty()` is already called for almost all tokenization.

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -84,38 +84,13 @@ var jsonata = (function() {
                 position++;
                 currentChar = path.charAt(position);
             }
-            // handle double-char operators
-            if (currentChar === '.' && path.charAt(position + 1) === '.') {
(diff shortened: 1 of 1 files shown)


andrew-coleman:
Looks good - thank you

andrew-coleman:
Oops - looks like this caused a regression when parsing name tokens that start with 'or' and 'in'.

E.g. customer.order  now throws a syntax error, because the 'or' in 'order' is pulled out as a separate token.  Test cases have been added to check this.  Will need to back this change out.

PR #44, opened 2017-04-14
Update tutorial.md

mw75:
Fixed headline markdown for better reading. I hope that was a bug and not a feature for any reason.

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -1,17 +1,17 @@
 # JSONata
 JSON query and transformation language
 
-###Introduction
+### Introduction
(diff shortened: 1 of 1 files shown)


mattbaileyuk:
@mw75 Thanks, well spotted. I think we got away with it until GitHub started tightening up markdown last month (https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown).

Nice to get this tidied up, although I think @andrew-coleman and I would like to deprecate that tutorial page in favour of the docs we've started building out at http://try.jsonata.org/ (GitHub pages-hosted)

PR #49, opened 2017-05-17
Base64 encode (btoa) and decode (atob) functions

peterbroadhurst:
These simple string<->base64 functions are helpful when constructing basic auth headers.
The code in this pull request provides trivial implementations that should work in both a browser (where btoa and atob) functions are available, and in a Node.js runtime where these are available via Buffer.

A further extension would be to support binary<->base64, but I'm not sure of:
- The right representation of binary in a JSONata structure (or if this is defined yet)
- How to distinguish the expected return value, when decoding a string. Should that be a different function name?

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -3120,6 +3120,41 @@ var jsonata = (function() {
         return result;
     }
 
+    /**
+     * Base64 encode a string
+     * @param {String} str - string
+     * @returns {String} Base 64 encoding of the binary data
+     */
+    function functionBase64encode(str) {
+        // undefined inputs always return undefined
+        if(typeof str === 'undefined') {
+            return undefined;
+        }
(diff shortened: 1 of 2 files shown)


andrew-coleman:
Looks good - many thanks.
If possible, could you add a few lines of documentation to docs repo, please?...
https://github.com/jsonata-js/jsonata-js.github.io/blob/master/pages/mydoc/string-functions.md

chrisbygrave:
```A further extension would be to support binary<->base64```

Worth noting that `atob` and `btoa` won't be suitable for this - they only support ASCII<->base64

peterbroadhurst:
Pull request on docs is here: https://github.com/jsonata-js/jsonata-js.github.io/pull/3

PR #50, opened 2017-05-17
Remove quotes around error message inserts

mattbaileyuk:
With JSONata 1.2 the error values are stringified before they are inserted into the error message; for some messages there were already quotes, so this is resulting in things like `'Expected '")"' before end of expression`.

These changes fix those, although there's still an issue for things like `Attempted to invoke a non-function. Did you mean $"uppercase"?`


andrew-coleman:
That last point is a problem - thanks for spotting that.  We'll need to fix that somehow.
Thanks

PR #51, opened 2017-05-19
ES5 versions

s100:
Recent changes in JSONata 1.2.x introduced the use of [ES6 generators](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/function*), but support for these is not universal, so an ES5-compatible version of JSONata is desirable.

* Added `babel-cli` as a development dependency in `package.json`.
* Added `babel-preset-env` as a development dependency in `package.json` - this provides a basic preset configuration for Babel, named "env".
* Added `.babelrc` to configure `babel` for this task. Note the use of the "env" preset, provided above.
* Added `babel-runtime` as a development dependency in `package.json`. This provides the transformation which turns ES6 generators into simple ES5 code.
* Added `babel-plugin-transform-runtime` as a development dependency in `package.json`. When ES6 code which makes use of generators is compiled to ES5, ordinarily the output ES5 code makes use of a global variable `regeneratorRuntime` to work... and just assumes that this global variable exists somehow. However, using this plugin means that the output ES5 code instead explicitly calls `var _regenerator = require('babel-runtime/regenerator');` and uses `_regenerator` to accomplish the same result.
* Added the "transform-plugin" to `.babelrc`.
* Added a new NPM task, "babel", which uses the configuration supplied in `.babelrc` to compiles `./jsonata.js` (ES6) to `./jsonata-es5.js` (ES5).

Unfortunately this leaves us in a situation where `jsonata-es5.js` has an external dependency on `babel-runtime/regenerator` - previously, JSONata had no external dependencies, and we want to keep it that way. So a few more steps were needed:

* Added `browserify` as a development dependency in `package.json`.
* Added a new NPM task, "browserify", which takes `jsonata-es5.js` as input and browserifies it in place. Note the use of the `--standalone jsonata` flag, which makes it so that the `jsonata` module is exposed as a global variable in the resulting UMD module.
* Since `browserify` pulls in all needed modules, and `jsonata` currently makes use of `Buffer`, we find that `browserify` pulls in a full browser implemention of `Buffer` along with `ieee754` and `base64`, which bloats the output. This is unnecessary, because `Buffer` is only used when `jsonata` is running on Node.js. Using `global.Buffer` instead of `Buffer` prevents this. (I wasn't able to find a `browserify` flag to accomplish this.)

And:

* Task "minify" minifies `jsonata.js` to `jsonata.min.js`. A new task "minify-es5" minifies `jsonata-es5.js` to `jsonata-es5.min.js`.
* `jsonata-es5.js` and `jsonata-es5.min.js` have been added to `.eslintignore` and `.gitignore`.

Possible future work:

* Separate out the JSONata source from its distribution, so that we can present `jsonata.js` as a properly bundled UMD module as well?
* Manually implement Base64 so that we don't have to rely on that unsightly `Buffer` hack and can bring JSONata back down to 0 external dependencies (at least in the ES6 case)?
* Is there a more elegant way to get that ES6 generator implementation in there, or replace it with something of our own?


andrew-coleman:
nice job! - many thanks

PR #53, opened 2017-05-25
Correctly handle null values when executed with a callback

sstone1:
The following error occurs when jsonata is invoked with a callback and an expression that returns a `null` value:

```
TypeError: Cannot read property 'then' of null
      at evaluate (jsonata.js:1377:58)
      at next (native)
      at evaluateStep (jsonata.js:1476:31)
      at next (native)
      at evaluatePath (jsonata.js:1429:38)
      at next (native)
      at evaluate (jsonata.js:1323:34)
      at next (native)
      at Object.evaluate (jsonata.js:4117:33)
      at test/async-function.js:20:14
      at jsonataPromise (test/async-function.js:19:12)
      at Context.<anonymous> (test/async-function.js:95:23)
```

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -1374,7 +1374,7 @@ var jsonata = (function() {
         }
 
         if(environment.lookup('__jsonata_async') &&
-          (typeof result === 'undefined' || typeof result.then !== 'function')) {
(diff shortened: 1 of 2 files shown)


andrew-coleman:
Thanks you!

PR #56, opened 2017-06-02
Identify generators

s100:
We were experiencing an issue where `jsonata-es5.min.js` was failing tests (`jsonata.js`, `jsonata.min.js` and `jsonata-es5.js` all worked okay). After much whittling down to a minimal failing case, it seems the problem was with how generator functions are identified in `isGenerator`. The approach currently in use works fine in ES6, but after the Babel step there are no native `GeneratorFunction` objects so a different approach as suggested [here](https://github.com/babel/babel/issues/2577) must be used.


s100:
Not sure why there are so many commits here, looking at the files changed you will see that this is only a small change.

Possible future work: adapt `jsonata-test.js` to test all four JSONata versions?

s100:
Uh, so the Travis builds fail on Node 5 and Node 4.4, which I expect is because those platforms don't support generators?

s100:
It seems I was mistaken in my earlier solution to this problem. In the general case [it is (deliberately) not possible](https://stackoverflow.com/questions/16754956/check-if-function-is-a-generator) to distinguish between a "generator function" and a conventional function. Generator functions as declared with `function*(){}` are in fact basically just syntactic sugar. As long as the value returned by a function is a generator, the function can be considered a "generator function"; furthermore, there is no reason why a function can't alternately return conventional values and generators.

So the correct solution to this problem is not to try to identify the function itself as a generator function but to identify the *return value* as a *generator*. In this latest commit I have modified `isGenerator` to do this. I verified that the result works in Node.js v4.4.7 (Windows x64) and I suppose Travis will shortly be able to tell us if this solution still works on later versions.

PR #62, opened 2017-06-27
Split large JSONata source file into smaller modules

blgm:
The JSONata source file is getting very long, and it would be better
broken down into smaller files.  This PR:
- Moves `/jsonata.js` to `/src` and splits off the numeric functions
- Uses ES2015 module format so that
  - Rollup can be used to join the files
- A CommonJS version is genereted in `/lib`
  - The tests are run against this version so that 100% code coverage can be achieved
- A UMD version is produced where the original source file was


xogeny:
Is this PR likely to be accepted?  It would make the code a lot more readable and might even lend some extensibility.

mattbaileyuk:
There was discussion on this away from this PR (something we should avoid...) which resulted in it not being merged at the time (IIRC, one concern was around not being able to just point directly at `jsonata.js` in GitHub which the exerciser uses for branch support) and now it would require some significant work to bring it back up to speed. 

I've separately been thinking about using `rollup` soon as another answer to https://github.com/jsonata-js/jsonata/pull/106 so I might do that bit soon, but as `jsonata.js` has continued to grow and grow I agree it would be good to get back to splitting up the file.

blgm:
This PR is now a little out of date, but I still think it's the right direction.  React 16 is now using `rollup`, so it's a well established packaging mechanism.  On reflection, I think this PR does too much in one go, and it could be broken down into incremental steps.

mattbaileyuk:
Some of the splitting up of modules was done separately for JSONata 1.5. I suspect there are still improvements to be made there, but I think we'd be better off deriving from master than continuing this branch.

Thanks for the contribution and for nudging us in the right direction.

Issue #73, opened 2017-08-30
Odd error message in tokenization


blgm:
Consider the following
```javascript
const jsonata = require('jsonata')
try {
  jsonata('$replace("foo", "o, "rr")')
} catch (e) {
  console.log(e.message)
}
```
The error message printed out is:
```
Expected ")", got "(name)"
```

Which is not what I would have expected.

PR #74, opened 2017-08-30
Write out token value, not internal type

blgm:
Fixes https://github.com/jsonata-js/jsonata/issues/73

The `S0202` message currently reports the internal token type, rather than the token value.  For some tokens that's ok (because they are the same), but for `name` and `literal` tokens at least, it's better for the user to see the values that they typed, rather than the confusing `(name)` and `(literal)`.


andrew-coleman:
Thank you!

Issue #84, opened 2017-10-24
Class in path


Allam76:
Hello Andrew,

If a ES6 class definition is in the path, it is not picked up.

In the code 1800:
```js
        } else if (input !== null && typeof input === 'object') {
```
This means that classes cannot be used. I propose the following addition:
```js
} else if (input !== null && ( typeof input === 'object' || isClass(input))) {
```
where isClass is defined as:
```js
    function isClass(func) {
        return typeof func === 'function' 
            && /^class\s/.test(Function.prototype.toString.call(func));
    }
```

/Martin

andrew-coleman:
JSONata operates on *JSON* input data.  JSON has no concept of classes.  As a general rule, if the input cannot survive a cycle of JSON.stringify() / JSON.parse(), then it's not a suitable input for JSONata.

Allam76:
I'm a bit sad to hear that. Jsonata could be used for so much more....

Anyway, could then the "mixin" concept be supported, so that I can override the functions I need with custom code without breaking every time you update the release?

andrew-coleman:
Closing this issue.  Sounds like @xogeny's implementation might address this which will hopefully be the basis for the next major version of JSONata.

PR #86, opened 2017-10-26
Support ES6 Class in path 

Allam76:
This small change supports the use of ES6 classes in jsonata path. Currently it can only support objects but the class is a special type of function.

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -1797,7 +1797,7 @@ var jsonata = (function() {
                     result.push(res);
                 }
             }
-        } else if (input !== null && typeof input === 'object') {
+        } else if (input !== null && ( typeof input === 'object' || isClass(input))) {
             result = input[expr.value];
         }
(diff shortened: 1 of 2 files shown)


mattbaileyuk:
Thanks for the pull request. Unfortunately it's currently failing the Travis builds due to some linting errors that ESLint has picked up on. You should be able to see these locally by running `npm t`. Could you please fix these issues and push a new commit to this fork+branch (then Travis will pick it up and automatically re-run); they look like they should be small things to resolve, but let us know if you need help with these.

mattbaileyuk:
This would appear to link to https://github.com/jsonata-js/jsonata/issues/84

andrew-coleman:
Thanks for the PR, but as per [comment](https://github.com/jsonata-js/jsonata/issues/84#issuecomment-340463867) in issue 84, I don't think this fits with the principle that JSONata only addresses JSON input structures.

PR #87, opened 2017-10-26
Mixins

Allam76:
This change support mixins in the evaluate name method. This way, jsonata can pick up info from other sources than the json input. Could be useful to fetch extra data from a db, for example.


andrew-coleman:
Closing as discussed in #132

PR #106, opened 2017-12-29
provide es5 as main entry point, es6 as module entry point

### LINKED ISSUE #105: Failed to minify in a webpack project
I'm trying to use jsonata in my front-end webpack project, importing it like this:

import jsonata from 'jsonata'

It works in a dev environment, but when I try to do a production build it fails with:

> Creating an optimized production build...
> Failed to compile.
> 
> Failed to minify the code from this file:
> 
>         ./node_modules/jsonata/jsonata.js:1438
> 
> Read more here: http://bit.ly/2tRViJ9

As per the referenced link, the main entry point is ES6 rather than ES5.

asfordmatt:
Fixes #105 

Signed-off-by: Matt Chapman <mpchapman@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "jsonata",
   "version": "1.4.1",
   "description": "JSON query and transformation language",
-  "main": "jsonata.js",
+  "module": "jsonata.js",
+  "main": "jsonata-es5.js",
   "homepage": "http://jsonata.org/",
   "repository": {
     "type": "git",


mattbaileyuk:
While I don't believe `module` is officially part of the npm spec for `package.json` it looks like it's getting enough use in the likes of `Webpack` and `rollup` to be worth including here to specify an ES6+ version.

However, I am a bit concerned about switching the default for `main` from the native ES6 code to the transpiled ES5 version, which was intentionally avoided when that version was introduced, but it might be time to revisit that. I'm mainly concerned about what that means for projects which import JSONata, in particular those like `node-RED` which use `require('jsonata')`.

mattbaileyuk:
Thinking about this more, I'm not as concerned about that change as I was previously, and think this would be good to have as I'm such other people picking this up for their front-end projects, whether they're using Webpack or otherwise, are likely to run into this.

PR #111, opened 2018-01-13
Establish some (common) VS code settings (for those using it)

xogeny:
This just initiates some project level settings for VSCode.  In this case, I just wanted to turn off `formatOnSave` because it might cause unnecessarily cosmetic changes if everybody formats the code according to different style guidelines.

Of course, this should not impact people who are not using VSCode.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    // To avoid different format settings from
+    // triggering cosmetic changes.
+    "editor.formatOnSave": false,
+}


xogeny:
I added my sign-off in the commit.  I hope that is how it is supposed to be done...

mattbaileyuk:
Personally I'm a little reluctant to include settings files for one editor as it opens the door to including files for all editors.

However, I'm not a VS Code user; is it particularly annoying to a developer without these settings?

xogeny:
Frankly, it doesn't matter to me one way or the other.  The issue here is that my normal user settings are such that VS Code will format code on save.  But you are running lint with rules that enforce spacing.  So if I have that setting active, it will immediately reformat the code and violate the lint rules.  Furthermore, if I edit any file, it will reformat it creating lots of cosmetic changes that could lead to unnecessarily merge conflict.  So I added this just to address those two problems.

An alternative, which I'm equally fine with, is to add `.vscode` to the `.gitignore` file so I can set these settings for my local working directory and not risk having them merged in with a commit.

Now, that being said I think the former is quite reasonable because a) VS Code is one of only a handful of really popular editor...so I don't think you will end up with that many editor configurations.  Also, it seems like there are a very small number of committers here too, thus reducing the odds further.  But, as I said before, I can live with either.  So if you don't want the editor configuration in here I'll change the PR to modify `.gitignore` instead.

mattbaileyuk:
Thanks for the clarification; it sounds like something most VS Code users of this repo would hit so happy to merge.

PR #112, opened 2018-01-13
Fix reversed test cases

xogeny:
Some test cases reverse the expected
result and the actual result.  Furthermore,
a handful of other tests use the term
expected to refer to an actual result.

These changes just clarify these cases.

Signed-off-by: Michael M. Tiller <michael.tiller@gmail.com>


mattbaileyuk:
Good catch, thanks 🙂

PR #113, opened 2018-01-14
New Language Neutral Test Suite

xogeny:
The majority of test cases have been refactored from Javascript code into a canonical test case form and stored as JSON.  This change makes it very easy for other implementation written in other languages to test compliance with `jsonata`.  It also helps organize the tests into categories.

Included in this commit is a markdown file documenting the data files present in the new test suite. 
 Also included is a mocha test that executes all of the tests by looping over all the provided test case data.

It is important to emphasize that no tests were removed.  They were only refactored into language neutral versions and a few lingering implementation specific versions.  Furthermore, these changes **preserve the current code coverage at 100% across the board**.

Signed-off-by: Michael M. Tiller \<michael.tiller@gmail.com\>


xogeny:
I don't expect this PR to be without controversy.  The very idea here is probably a big change in how you are used to doing things.  To me, for `jsonata` to really gain wider attention it is essential that other implementations can easily check conformance with the "reference implementation".  One could, of course, do a "one off" port of the test suite and export that as JSON.  But it would almost certainly become out of date with the reference implementation (and probably quickly).  Make sure that the reference implementation uses this language neutral data **itself** means that every implementation is on an equal footing which should reassure potential developers in other languages.

Personally, I think the idea of `jsonata` is long overdue and this is just an attempt to encourage developers to implement it.  So I hope this PR will be accepted and I'll gladly answer any questions about it.

blgm:
I think this should be the strategic direction.  I had a go at this myself some time ago (#26), but this is a more complete approach.

xogeny:
@blgm Doh! I didn't even see #26!  I really should have researched this a bit more.  I probably should have started there.

However, I notice that that pull request languished without comment.  So I wonder if this one will be in the same boat? 😞

mattbaileyuk:
@xogeny This looks great, and agree it's the way to go with Java and Go implementations of JSONata on the way. I like the format of the tests too, and all looked nice when running locally.

There's unfortunately a merge conflict due to that test fix of yours I merged the other day. However, I'm also reluctant to merge this at the moment as I'm aware that @andrew-coleman was planning to imminently merge in the `singleton` branch to `master` which has some additional test cases in `jsonata-test.js`. Is that OK with you?

xogeny:
I can totally rebase all this stuff as needed.  If I understand the situation correctly, I'd need to rebase to take into account my previous PR that corrected the reversed and expected values and also rebase on top of whatever @andrew-coleman merges from the `singleton` branch, correct?

I can absolutely do that.  Seems like this is all contingent on @andrew-coleman's merge.  So should we plan to resolve all this once that merge happens?

xogeny:
OK, I updated this pull requests rebasing it off of the `singleton` branch **and** improving the structure of the test suite data (IMHO).

All tests pass (including the new tests added on the `singleton` branch) and we still have 100% coverage.

Interestingly, no new tests were actually needed to get 100% test cases, just updated answers...but I added the new tests all the same.

xogeny:
FYI, I don't quite understand why the branch is marked as having conflicts.  I've removed `test/jsonata-test.js` and replaced it with two other files.  It should merge perfectly fine (since the deletion and additions are all in the commit).  So just be aware that this *should* be clean.  Feel free to try it and forward any problems to me.  But I rebased everything from `origin/master` so it should work just fine.

xogeny:
I'm closing this pull request and opening a different one (#122).  Something isn't right with this branch so I've just created a new one (that incorporates the rebase and the restructuring).  So just ignore this PR in favor of #122.

PR #118, opened 2018-01-17
Changes to how 'singleton sequences' are handled

### LINKED ISSUE #93: Clarification on JSONata handling of (nested) singleton arrays
I am working on a Java implementation of the JSONata engine, and I'm struggling to understand JSONata's treatment of singleton arrays...

From http://docs.jsonata.org/complex.html:

_Within a JSONata expression or subexpression, any value (which is not itself an array) and an array containing just that value are deemed to be equivalent._

Firstly, if so - why does the following hold?

`[1] = 1` => `false` _(should this not be `true`?)_

So, outside of the '=' operator itself, I suspect the singleton/primitive equivalence is effectively achieved by "unwrapping" singleton arrays as they are referenced by expressions.

I see the following behaviour:

`{'a':        1     }.a` =>      `1`
`{'a':       [1]    }.a` =>      `1`
`{'a':      [[1]]    }.a` =>     `1`
`{'a':     [[[1]]]    }.a` =>   `[1]`
`{'a':    [[[[1]]]]    }.a` => `[[1]]`

so it seems to me that JSONata is applying (at most) two levels of flattening to nested singleton arrays pulled out of an associative array via a path. Why 2? Why does it not completely flatten the output array in cases like this?

Next, if I wrap the statements above in an array constructor, I see the the following:

`[{'a':        1     }.a]` =>      `[1]`
`[{'a':       [1]    }.a]`=>       `[1]`
`[{'a':      [[1]]    }.a]` =>     `[1]`

these results seem reasonable (given that the embedded statement returns `1` in those cases), however:

`[{'a':     [[[1]]]    }.a]` =>   ` [1]`

should this not be `[[1]]`? (given than the statement inside the array constructor evaluates to `[1]`?

andrew-coleman:
Largely internal changes to how singleton sequences are stored and passed around in the evaluator.  A few tests modified to accommodate the change of behaviour - mostly edge cases.
This PR should resolve #93

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -1881,7 +1881,6 @@ var jsonata = (function() {
         } else if (input !== null && typeof input === 'object') {
             result = input[expr.value];
         }
-        //        result = normalizeSequence(result);
         return result;
     }
 
--- a/jsonata.js
+++ b/jsonata.js
@@ -1453,7 +1453,6 @@ var jsonata = (function() {
         switch (expr.type) {
             case 'path':
                 result = yield * evaluatePath(expr, input, environment);
-                // result = normalizeSequence(result, expr.keepSingletonArray);
                 break;
             case 'binary':
                 result = yield * evaluateBinary(expr, input, environment);
@@ -1522,7 +1521,6 @@ var jsonata = (function() {
 
         if (expr.hasOwnProperty('predicate')) {
             result = yield * applyPredicates(expr.predicate, result, environment);
-            // result = normalizeSequence(result);
 
         }
         if (expr.hasOwnProperty('group')) {
@@ -1591,10 +1589,6 @@ var jsonata = (function() {
         return arr;
     }
 
-    // Array.prototype.value = function() {
-    //     return this;
-    // };
-
     /**
      * Evaluate path expression against input data
      * @param {Object} expr - JSONata expression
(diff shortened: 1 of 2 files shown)


mattbaileyuk:
@andrew-coleman Are there any readme changes which should come in because of this? Might only be the docs in the other repo.

andrew-coleman:
I think readme is unaffected.  Should definitely review docs

jsonata.js
```diff
@@ -1516,7 +1522,7 @@ var jsonata = (function() {
 
         if (expr.hasOwnProperty('predicate')) {
             result = yield * applyPredicates(expr.predicate, result, environment);
-            result = normalizeSequence(result);
+            // result = normalizeSequence(result);
```
mattbaileyuk: Commented out code to remove

jsonata.js
```diff
@@ -1446,8 +1452,8 @@ var jsonata = (function() {
 
         switch (expr.type) {
             case 'path':
-                result = yield * evaluatePath(expr.steps, input, environment);
-                result = normalizeSequence(result, expr.keepSingletonArray);
+                result = yield * evaluatePath(expr, input, environment);
+                // result = normalizeSequence(result, expr.keepSingletonArray);
```
mattbaileyuk: Commented out code to remove

jsonata.js
```diff
@@ -1528,9 +1534,67 @@ var jsonata = (function() {
...
+                    }
+                };
+            }
+        });
+        return arr;
+    }
+
+    // Array.prototype.value = function() {
+    //     return this;
+    // };
```
mattbaileyuk: Commented out code to remove

jsonata.js
```diff
@@ -1580,48 +1648,78 @@ var jsonata = (function() {
...
+    //         if(keepSingleton || sequence.keepSingleton) {
+    //             result = sequence;
+    //         } else {
+    //             result = sequence[0];
+    //         }
+    //     } else if (sequence.length > 1) {
+    //         result = sequence;
+    //     }
+    //     return result;
+    // }
```
mattbaileyuk: Commented out code to remove

jsonata.js
```diff
@@ -1580,48 +1648,78 @@ var jsonata = (function() {
...
+                //         console.log('ouch!');
+                //     }
+                // res is a sequence - flatten it into the parent sequence
+                // Array.prototype.push.apply(result, res);
+                // res.forEach(function (innerRes) {
+                //     //if (typeof innerRes !== 'undefined') {
+                //     result.push(innerRes);
+                //     //}
+                // });
+                // }
```
mattbaileyuk: Commented out code to remove

jsonata.js
```diff
@@ -1859,8 +1963,8 @@ var jsonata = (function() {
             });
         }
 
-        result = normalizeSequence(results);
-        return result;
+        //        result = normalizeSequence(results);
```
mattbaileyuk: Commented out code to remove

mattbaileyuk: Just this one left... 😄

xogeny:
Excited to see these changes getting merged. 👍

PR #131, opened 2018-01-22
Initial v2 refactor

xogeny:
Lots of changes here.  First and foremost, this organizes the code into es6 modules.
I tried to hit the big functional bigs: tokenizer, parser, evaluator, functions and separate
them out.  The parser was particularly involved so I actually further decomposed that
so that files were at least a reasonable size.  I don't have particularly strong feelings
about the taxonomy withing the parser package, just trying to get things to a
manageable state.

There are several other big changes worth noting here in the tooling.  First, I migrated
the code to TypeScript.  I think TypeScript is a particularly good fit for AST related
work because of the tagged union support.  But it also really helps with refactoring
so I think it will be pretty useful moving forward.  It all compiles out to es6 and es5
and can be bundled with rollup.  So it doesn't impact *deployment* at all, just
development.

I moved the tests to `jest`.  I find Jest is a little easier to configure out of the box
and things like coverage are well supported.

I used `yarn` to create a `yarn.lock` file.  We could also use `npm` `v5`, but since
`jsonata` currently supports Node `v4`, and `npm` `v5` only ships with Node v7+.
Bottom line is that `yarn` (or `npm5`) give much better reproducibility on installs.

Even with all these changes, the test suite still passes 100% of tests with 100% coverage!


andrew-coleman:
Looks like jest doesn't support node v4

mattbaileyuk:
They dropped support at the end of last year in version 22; we could lock to a backlevel (21) but not sure how long that would be for. Might need to decide on the future of node v4 support in JSONata as it drops completely from any further fixes (even security) in April. Might depend on `node-red`, though.

xogeny:
OK, it turns out that downgrading to `jest` `v21` seems to work.  I'll push out a new version of the branch that should hopefully address this.

xogeny:
...and it seems that downgrading wasn't sufficient.

DaniGuardiola:
What is the status of this? Need any help?

mtiller:
I actually got the code in reasonable shape.  I believe the entire original test suite passes in my version along with probably a few additional tests I added.  You can find it on the `v2-refactor` branch in [my fork](https://github.com/mtiller/jsonata).

I do not plan to continue that work.  Ultimately, the `jsonata` project operates under various (reasonable) constraints that just made it difficult for me to take the code where I wanted it to go.  But hopefully somebody can incorporate some of the ideas I pursued there.  A bit part of my refactoring was:
- Typing all AST node types to allow type checking during interpretation
- Using multiple files rather than just one
- Using TypeScript
- Changing the "working value" model for evaluation so that it used a "boxed" value rather than trying to attach various attributes internal attributes to the value itself.  There may very well be performance degradation in what I did, FYI.

The biggest issues I found were several cases of what I would call "inconsistent semantics" where the evaluation of a node was context sensitive.  Another issue I had was related to how to handle sync and async cases without having to rewrite massive amounts of code.  I kind of managed this I think, not sure.  I was shooting for some abstractions where the fundamental evaluation would always be the same, but the abstractions would take care of issues like sync vs. async.  This is slightly related to another goal I had which was to be able to apply `jsonata` expressions to databases.  The issues with sync vs. async also come into play there.

I also found the way that lambdas were being evaluated hard to follow and abstract.

If anybody has any questions, I'd be happy to answer them.

PR #141, opened 2018-01-31
Update regex in toMillis to allow more complete range of ISO 8601 compatible dates

### LINKED ISSUE #140: ISO 8601 check for toMillis is too strict
The ISO 8601 date check for the `toMillis()` function is too strict and only accepts dates in the format `2018-01-31T17:39:00.000Z` or `2018-01-31T17:39:00.000+00:00` rather than the wider range of valid dates which is possible such as `2018-01-31` or `2018-01-31T17:39:00.000+0000` (the latter of which is where I stumbled upon this issue).

mattbaileyuk:
Fixes #140 by adjusting the regex to a complete range of ISO 8601 dates which `Date.parse()` is subsequently happy with. Also updated one of the `toMillis()` test cases to use a date format which is now compatible through this.


mattbaileyuk:
Good point; I found those cases in a search and didn't spot where they were located.

In fact, I'm glad I added some additional test cases for other formats that the regex is designed to match as I discovered that `Date.parse()` in Javascript (which `toMillis()` is using) is stricter with the ISO 8601 dates it supports, although it's still quite a lot more than the old regex allowed (e.g. `20110105` isn't supported by that method, but `2011-01-05` is - although the latter was previously rejected).

So I will need to scope down the regex string, or possibly rely on `Date.parse()` giving back `NaN` for anything it doesn't support, but that feels a bit too Javascript-specific when there are implementers of JSONata in other languages now.

mattbaileyuk:
Because of the `Date.parse()` limitations I've narrowed the regex down to supporting the following:
- `2018-01-01T01:02:03.456Z` (as before)
- `2018-01-01T01:02:03.456+00:00` (as before)
- `2018-01-01T01:02:03.456+0000`
- `2018-01-01T01:02:03.456`
- `2018-01-01T01:02:03`
- `2018-01-01`
- `2018-01`
- `2018`
which are all ISO 8601 valid, and supported by `Date.parse()`. There are technically more ISO 8601 dates that could be supported, but this seems a sufficient set to support a wide set of use cases.

There are additional tests to cover a number of these.

jsonata.js
```diff
@@ -4653,7 +4653,8 @@ var jsonata = (function() {
     }
 
     // Regular expression to match an ISO 8601 formatted timestamp
-    var iso8601regex = new RegExp('^\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\d\\.\\d+([+-][0-2]\\d:[0-5]\\d|Z)$');
+    // Uses https://www.myintervals.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/ but removes the "week number" format not supported by Date.parse
```
mattbaileyuk: In hindsight I should have removed this comment as I ended up crafting my own regex...

andrew-coleman: raise another PR :)

PR #175, opened 2018-03-22
throw error if multiple group-by expressions evaluate to same key

### LINKED ISSUE #163: Semantic Issue in Object Grouping
I noticed, while reviewing the evaluation code, that there is what I think is an incorrect assumption.  The evaluation for objects appears to work as follows:

We loop over each input value.  We evaluate the `key` expression in the context of that input value.  We then append that input value to the list of values we will use later for that key (more on that in a second).  We *also* note the expression associated with that key and record that as well so we can evaluate that value expression in the context of each of the inputs we are storing away.  But here is the problem...the value expression isn't always the same but we only store the first one we run into.  I'm pretty sure this isn't correct and it creates some odd issues.  For example, imagine our input (for all cases I'm going to consider) is:

```json
[
  {
    "key": "foo",
    "value": 5
  },
  {
    "key": "bar",
    "value": 10
  }
]
```

Now if I evaluate this: `{ key: value }`, I get pretty much what you might expect, *i.e.,* `{ "foo": 5, "bar": 10 }`.

So far so good.  If I evaluate `{ key: (value+1) }`, I also get pretty much what I would expect: `{ "foo": 6, "bar": 11 }`.

Now how about if I evaluate: `{ key: value, key: value }`.  Now we have **overlapping values for the key expressions**.  That's actually ok and it works (and I think it is reasonable to allow overlapping key expressions, BTW).  It evaluates to:

```json
{
  "foo": [
    5,
    5
  ],
  "bar": [
    10,
    10
  ]
}
```

I'm no so sure this is the most intuitive result.  Some might expect the second to *overwrite* the first in this case (*i.e.*, some might expect `{ "foo": 5, "bar": 10 }`), but it doesn't.  Instead, it appends the results.  But at least we get an answer that we can explain even if it might seem unintuitive.

But now things start to misbehave.  Imagine I evaluate this instead: `{ key: value, key: value+1 }`.  I get the same answer as for `{ key: value, key: value }`.  This is because `value` is tucked away as the expression associated with `"foo"` and `"bar"` because it is the first one encountered, so `value+1` **is never evaluated**.  But things can actually get worse.  If you switch the order and try to evaluate `{ key: value+1, key: value }` you actually get an error that it cannot evaluate `value` because it isn't a number.

Now, one thing you can do is change the semantics slightly so you actually evaluate the `value` expressions immediately in the context of each input value (rather than storing them up).  This actually makes sense for the `value+1` case and seems reasonably intuitive.  Said another way, it may be more intuitive to do the evaluate this way because I suspect it is what people would expect.  But, this will actually breaks one of the test cases, *i.e.,* the containing the expression:

```Account.Order.Product{$string(ProductID): (Price)[0]}```

Because the `[0]` in there become meaningless because `Price` always evaluates to a scalar rather than an array.  So that approach would change the semantics slightly.  Presumably the `[0]` is there specifically to grab the first value.

The alternative is to append not just each successive input value associated with each key but also each unique value expression as well (instead of just the first one).  I suspect you'd probably want this alternative but I wanted to thoroughly explain the problem and collect feedback to be sure.

To be honest, implementing these evaluation semantics is a bit complicated.  Furthermore, that error for `value+1` is pretty confusing and unintuitive.  So before working on implementing anything more (I've implemented a couple different approaches already), I thought it would be good to discuss what is most intuitive.

Any comments?

### PR DESCRIPTION
**Pull request by andrew-coleman**
New dynamic error created which is thrown if multiple clauses within a 'group-by' expression `{...}` evaluate to the same key.  Previously, unexpected results would get returned.

Resolves #163 as agreed with @xogeny

### CODE REVIEW

File: jsonata.js [hunk 1]
```diff
@@ -2219,6 +2224,20 @@ var jsonata = (function() {
...
+                            code: "D1009",
+                            stack: (new Error()).stack,
+                            position: expr.position,
+                            value: key
+                        };
+                    }
+                } else {
+                    owners[key] = pairIndex;
+                }
                 var entry = {data: item, expr: pair[1]};
```
xogeny: Note that instead of storing `expr` here, you could simply store the `pairIndex` and then lookup the expression later as `expr.lhs[pairIndex][1]`.  The point being that what you are trying to do here is avoid having to compare expressions for equality (which isn't feasible) so you are instead storing the index of the pair whose value expression you are using.  But you could use that `pairIndex` everywhere and not store the expression.  Then, you'd only need to see if the pairIndex matched when you attempted to append a new item to `data`.

Said another way, don't store `expr` in the `entry`, store the `pairIndex`.  Then if you find a key already exists, you won't even need the `owners` object, just check to make sure that you have the same `pairIndex` as you did when the entry was created.

Does that make sense?

xogeny: Alternatively, you can store `expr` **and** `pairIndex`.  Same effect, you don't need `owners` at all.

andrew-coleman: That's true.  I'll make that change. Thanks!

PR #189, opened 2018-04-05
Move reverse test cases into their own directory

deepilla:
This PR moves the test case files for the `$reverse()` function into a new directory. They used to be in the same directory as the `$each()` tests.


mattbaileyuk:
Thanks for spotting this and fixing it.

Issue #192, opened 2018-04-09
Functions for dealing with char codes


brodo:
I am using node-red to talk to IoT devices via modbus and so far I could use jsonata for all data transformation necessary. However, today I needed to convert between characters and their character codes. JavaScript has the `String.fromCharCode(INT)` and `STRING.charCodeAt()` functions for that. It would be nice to expose these to jsonata. Maybe as `$charCode()` and `$fromCharCode()`.  I am willing to do a PR for that. However, before I put any effort into it I would like to know if you guys are interested in it.

brodo:
@michaelthoward I changed it in the title now.

andrew-coleman:
Thanks for the proposal.  I think there might be a common theme here with some of the discussion in #80 (Proposal: Function(s) for casting string-encoded binary data to numeric values) and #171 (proposal: $string should work with Node.js' Buffer).  I'd like to consider all of these together with your proposal #200 (Proposal: Functions for doing binary and, or, xor, not and shifting) to make sure we deliver a set of consistent capabilities.  As @xogeny suggests, this might be a good candidate for an extension package designed specifically for these IoT type use-cases.

Given that, I'd like to put your pull requests #198 and #199 on hold for a bit.

PR #199, opened 2018-04-14
Add new function: $fromCharCodes([int])

brodo:
This is the second function mentioned in #192 
It converts an array of char codes into a string.


brodo:
I'm spring-cleaning my Github account, closing down old and stale PRs for more clarity.

PR #209, opened 2018-04-19
Move clone test cases to the JavaScript test suite.

deepilla:
This PR removes all tests that use the `$clone()` function from the language-agnostic test suite and replaces them with JavaScript tests.

The `$clone()` function was added to jsonata-js solely as a workaround for Node-RED. It's not required by other implementations and therefore should not be included in the standard tests.


mattbaileyuk:
@deepilla Thanks for making this change

PR #216, opened 2018-05-04
Fix #204

### LINKED ISSUE #204: $substring length parameter no longer works with negative start positions
I've just noticed that the `$substring()` function isn't handling its length argument correctly when the start argument is negative. For example, this expression correctly evaluates to `"world"`:

```
$substring("Hello world", -5, 5)
```

But so do all of these expressions:

```
$substring("Hello world", -5, 4)    // should return "worl"

$substring("Hello world", -5, 3)    // should return "wor"

$substring("Hello world", -5, 2)    // should return "wo"

$substring("Hello world", -5, 1)    // should return "w"

$substring("Hello world", -5, 0)    // should return ""

$substring("Hello world", -5, -1)   // should return ""
```

I'm guessing that this bug was introduced in the recent fix for [UTF-16 surrogate pairs](https://github.com/jsonata-js/jsonata/issues/156).

brodo:
This PR fixes #204 .

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -3112,6 +3112,9 @@ var jsonata = (function() {
 
         var strArray = Array.from(str);
         if(typeof length !== 'undefined') {
+            if(start < 0){
+                start = strArray.length + start;
+            }
             var end = start >= 0 ? start + length : strArray.length - start + length;
             return strArray.slice(start, end).join('');
         }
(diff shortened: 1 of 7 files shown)


andrew-coleman:
Many thanks for looking at this issue, but unfortunately this PR couldn't be merged because it reduced test coverage (and hence rejected by travis).  This has now been addressed by PR #225

PR #219, opened 2018-05-10
async driver for full test suite, and fix all failing cases

### LINKED ISSUE #217: $filter returns undefined in async mode
I am not able to use the filter function in async mode. It always returns undefined ie.
```
const out = await new Promise((resolve) => {
  const expression = jsonata('$filter(data, function($d) { true })');
  expression.evaluate(input, {}, (err, value) => {
    console.log('ERROR?', err); // null
    resolve(value); // undefined
  });
});
```
data is a valid array and the expression `data` returns the array. Also using `$filter` in synchronous mode works as expected (using version 1.5.3).

andrew-coleman:
Created a test driver to run the test suite in async mode
Fixed all failing test cases.  These all involved functions which needed to asynchronously invoke the evaluator (such as the higher-order functions, the sort syntax & function chaining operator)

resolves #217

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@ jsonata-es5.js
 jsonata-es5.min.js
 node_modules/
 npm-debug.log
+package-lock.json
 .idea/
--- a/jsonata.js
+++ b/jsonata.js
@@ -4556,12 +4556,6 @@ var jsonata = (function() {
         } else {
             comp = function* (a, b) {
                 return yield * apply(comparator, [a, b], null);
-                // // returns a generator - so iterate over it
-                // var comp = it.next();
-                // while (!comp.done) {
-                //     comp = it.next(comp.value);
-                // }
-                // return comp.value;
             };
         }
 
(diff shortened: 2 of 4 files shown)


xogeny:
As far as I can tell (correct me if I'm wrong), the change here is to pass a callback to JSONata.  I think more rigor should be employed here.  For example, passing the dataset as a Promise and the bindings as Promises as well.  Ideally, also including variations that do the same with generators.

In analyzing this myself, my thinking was that there are essentially 3 fundamental sources for values in JSONata (again, correct me if I'm wrong).  Those are input data set, bindings and results from function evaluations.  It is those cases where we have to check whether something is a "normal" value, a Promise or a generator.  So my point here is that it seems that your change looks at the function return values.  But it doesn't exercise the other two cases.  Am I correct?

andrew-coleman:
This change doesn't alter the externals for triggering async mode, it merely tests it properly and fixes the failing cases.  I agree that we could modify the API to allow the passing in of promises as you suggest, but this goes beyond the scope of this bug fix.

PR #224, opened 2018-05-22
add support for custom matcher functions

### LINKED ISSUE #213: Defining functions for use with $match etc.
As discussed in [this thread](https://github.com/jsonata-js/jsonata/issues/177), the functions `$match`, `$contains`, `$split` and `$replace` all take a function that returns the matches of a particular pattern in a string. Regular expressions are one type of matching function but users can provide their own. I tried this out (to check that jsonata-go is doing the right thing) and found it quite tricky to get working.

I expected this expression to return one match:

```
(
    $f := function($arg) {
        {
            "match": "ab",
            "start": 0,
            "end": 2,
            "groups": [],
            "next": function() { nothing }
        }
    };

    $match("abracadabra", $f)
)
```

But instead it returned an error: `t is not a function`. So I re-ran the expression in the JavaScript console and got the slightly longer error: `Uncaught TypeError: regex is not a function` (`regex` is the name of the argument to `functionMatch`).

Ok, so lambdas can't be used as matching functions. Not a problem. I rewrote the function in JavaScript, added it to JSONata with `registerFunction`, and used that as the second argument to `$match`. Same error!

Looking at the code, I saw that `functionMatch` calls its `regex` argument directly. This means that `regex` has to be a native JavaScript function. But `registerFunction` wraps the function passed to it in an object. I tried using `assign` instead of `registerFunction` and that fixed the problem.

I have one observation and one question about this:

1. Functions added via `registerFunction` should work. It doesn't make sense for the same matching function to be valid when bound with `assign` and invalid when bound with `registerFunction`.

2. Is there a reason why lambdas can't be used as matching functions or is it just an oversight?

p.s. I only tried `$match` and `$contains` but I can see from the code that `$split` and `$replace` will have the same issues. They all call the matching function directly.

andrew-coleman:
The four functions `$match`, `$contains`, `$split` and `$replace` can take a regex argument.  The regex syntax `/.../` is in fact just some syntactic sugar for a higher-order function that generates a 'matcher' function.  This matcher function, when invoked on a string, will return a well defined structure (which I haven't documented yet :blush:) which describes the match(es).  It was designed so that these four functions could work with any pluggable matching scheme, not just the built-in regex, provided the matcher returned the correct object structure.  That's the theory, but I never got round to testing it.  This PR fixes that.  Custom matchers can now be implemented in jsonata as lambdas or as external javascript functions.

Resolves #213

### DIFF
--- a/test/run-test-suite-async.js
+++ b/test/run-test-suite-async.js
@@ -58,7 +58,7 @@ function jsonataPromise(expr, data, bindings) {
 describe("JSONata Test Suite - async mode", () => {
     // Iterate over all groups of tests
     groups.forEach(group => {
-        let casenames = fs.readdirSync(path.join(__dirname, "test-suite", "groups", group));
+        let casenames = fs.readdirSync(path.join(__dirname, "test-suite", "groups", group)).filter((name) => name.endsWith(".json"));
         // Read JSON file containing all cases for this group
         let cases = casenames.map((name) => readJSON(path.join("test-suite", "groups", group), name));
         describe("Group: " + group, () => {
@@ -66,6 +66,11 @@ describe("JSONata Test Suite - async mode", () => {
             for (let i = 0; i < cases.length; i++) {
                 // Extract the current test case of interest
                 let testcase = cases[i];
+
+                // if the testcase references and external jsonata file, read it in
+                if(testcase['expr-file']) {
+                    testcase.expr = fs.readFileSync(path.join(__dirname, "test-suite", "groups", group, testcase['expr-file'])).toString();
+                }
 
                 // Create a test based on the data in this testcase
                 it(casenames[i]+": "+testcase.expr, function() {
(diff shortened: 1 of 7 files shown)


test/run-test-suite-async.js
```diff
@@ -67,6 +67,11 @@ describe("JSONata Test Suite - async mode", () => {
                 // Extract the current test case of interest
                 let testcase = cases[i];
 
+                // if the testcase references and external jsonata file, read it in
```
mattbaileyuk: small typo here: `and` instead of `an`

test/run-test-suite.js
```diff
@@ -49,6 +49,11 @@ describe("JSONata Test Suite", () => {
                 // Extract the current test case of interest
                 let testcase = cases[i];
 
+                // if the testcase references and external jsonata file, read it in
```
mattbaileyuk: and here

Issue #156, opened 2018-02-19
$length and other string functions return misleading results for UTF-16 surrogate pairs


deepilla:
If you request the length of the smiley face emoji, JSONata returns 2 instead of 1.

    $length("😂")  // returns 2

The same thing happens with any of the more exotic Unicode symbols (emoji, obscure mathematical and musical notation, anything outside the [BMP](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) basically).

This is because JavaScript uses UTF-16 under the hood and needs two characters (known as a [surrogate pair](https://en.wikipedia.org/wiki/UTF-16#U+10000_to_U+10FFFF)) to represent non-BMP symbols. The equivalent JavaScript code also returns 2.

    "😂".length;  // also returns 2

This is technically correct -- surrogate pairs _do_ have two characters. But it's not very intuitive. It's also dependent on implementation language. I'm working on a Go implementation of JSONata and it doesn't have this issue (because Go uses UTF-8).

Some other expressions that return results you might not expect:

    $substring("😂", 1)  // returns an invalid character instead of an empty string

    $substring("😂", 0, 1)  // returns an invalid character instead of the emoji

    $pad("😂", 2)  // returns the emoji unpadded

    $split("😂", "")  // returns an array of two invalid characters

Note: all of these examples were run in JSONata Exerciser. I'm not sure what version of JSONata that uses.

PR #225, opened 2018-05-22
$substring should handle negative start position and length

### LINKED ISSUE #204: $substring length parameter no longer works with negative start positions

andrew-coleman:
The fix for #156 introduced a regression into the `$substring` function when dealing with negative values of start parameter (count from end of string).  This was caused by a miscalculation of the end position.  Also added checks to prevent start position and length from going below zero, in line with the previous behaviour.

Resolves #204

### DIFF
--- a/jsonata.js
+++ b/jsonata.js
@@ -3087,9 +3087,16 @@ var jsonata = (function() {
             return undefined;
         }
 
+        if(str.length + start < 0) {
+            start = 0;
+        }
+
         var strArray = Array.from(str);
         if(typeof length !== 'undefined') {
-            var end = start >= 0 ? start + length : strArray.length - start + length;
+            if(length <= 0) {
+                return '';
+            }
+            var end = start >= 0 ? start + length : strArray.length + start + length;
             return strArray.slice(start, end).join('');
         }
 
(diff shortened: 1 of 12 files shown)


deepilla:
I don't think this is quite fixed yet. The start position check uses the length of the string (i.e. the number of bytes) rather than the number of characters in the string. This can produce incorrect results when the input contains surrogate pairs, e.g.

```
functionSubstring("😂😁😀", -4, 1)    // returns "" instead of "😂" 
functionSubstring("😂😁😀", -5, 1)    // returns "😂😁" instead of "😂"
```

andrew-coleman:
Good spot. Thanks!
Hopefully now fixed.

Issue #214, opened 2018-04-24
create-react-app - 'npm run build' - fails


moccadroid:
When I use jsonata with create-react-app it fails when running 'npm run build' with the following error:

    Failed to minify the code from this file:
    ./node_modules/jsonata/jsonata.js:1464
    Read more here: http://bit.ly/2tRViJ9

according to the link I should open up a ticket for this on the dependency's issue tracker :)
please fix this? :D

xogeny:
I suspect this is actually a duplicate of #211.  You should be able to do an `npm run start` no problem, right?  It is just the build that is failing.  This is because the minifier doesn't support es6 syntax (which seems long over due, to me...but I digress).

s100:
So, `package.json`'s `"main"` property points to `jsonata-es5.js` (pure ES5) whereas its `"module"` property (intended for consumers which understand ES6 modules) points to `jsonata.js` (ES6 with `yield*` and suchlike all over the place). It's disappointing that support for ES6 modules is so out of whack with support for ES6 generators, but the bottom line is I think we need to move the *original* ES6+generators source to a separate file, and transpile away that generator syntax when constructing `jsonata.js` for consumers.

xogeny:
Just a comment...in my TypeScript based implementation there is no generator syntax.  Now, part of the reason is that my goal was to separate out the semantics of JSONata from how values were resolved.  So my current version only supports synchronous evaluation.  My goal is to actually incorporate async (promises, generators and/or observables) but I wanted to handle that as a separate concern since there are so many potential ways that values could be resolved and I didn't want to repeat all the semantics for each case.

Also note that because the code base is in TypeScript, it is possible to transpile to both es5 or es6 from the same (TypeScript) code base.

Unfortunately, that work is still ongoing.  But I just wanted to point out my goals.

andrew-coleman:
I've done a quick test to confirm that this works with jsonata v1.6.1.
The following create-react-app code builds and runs without error:

```
import React, { Component } from 'react';
import logo from './logo.svg';
import './App.css';
import jsonata from 'jsonata';

class App extends Component {
  render() {
    return (
      <div className="App">
        <header className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
            <p>JSONata: {jsonata('$formatInteger(1234, "w")').evaluate({})}</p>
          <p>Edit <code>src/App.js</code> and save to reload.
          </p>
          <a
            className="App-link"
            href="https://reactjs.org"
[...]

Issue #211, opened 2018-04-20
Failed to uglify


zambrey:
UglifyJS fails when trying to webpack a pacakge using jsonata with this error "UglifyJs Unexpected token: operator (*)".
Saw a similar issue reported in another repository, https://github.com/hyperledger/composer/issues/1062.
Version 1.1.1 seems to work for me like mentioned in that issue. The version that failed was the latest one, 1.5.3.

xogeny:
My guess is that some ES6 construct is getting in there.  I don't know much about the build process for `v1.5.x`, but presumably babel should be taking care of processing the ES6 into ES5.  But @michaelthoward mentioned what is probably the same issue in #202.

As I mentioned in other places, I'm working on a TypeScript implementation of JSONata that can be transpiled into either ES5 or ES6, but it is still in development so it doesn't help for now.

michaelthoward:
I also suspect that the issues is an ES6 construct. 
I also experienced failure on a * token when trying to load 1.5.3 under Java 8 nashorn JavaScript ... but have no trouble with 1.1.1 under Java 8 nashorn.

s100:
@zambrey Can you provide a minimal series of steps to reproduce this problem? Does UglifyJS tell you exactly where (on what line, in which source file) it saw this unexpected token? If so, it would be really helpful to know.

s100:
Circumstantial evidence makes me think that UglifyJS is following `package.json`'s `"module"` property to reach `jsonata.js` (full of generator syntax) instead of following `"main"` to reach `jsonata-es5.js` (pure ES5). We need to split our generator-heavy source out from the built `jsonata.js`, which can be ES6 but should have no generator syntax.

Issue #202, opened 2018-04-15
Get JSONata to operate under Java 8 nashorn JavaScript ES5 engine


michaelthoward:
Would be good to get JSONata (currently 1.5.3) to operate under the Java 8 nashorn ES5 engine. 

**Background**

The Java 8 nashorn JavaScript engine supports ES5 == ECMA Script 5. 
JSONata 1.1.1 can be easily run under Java 8 nashorn. 
JSONata 1.2 began using ES6 == ECMA Script 6 features. [See this thread.](https://stackoverflow.com/questions/48312771/using-jsonata-in-java-throws-exception/49587106#49587106)

The babel transpiler generates jsonata-es5.js from ES6 source. However, the transpiled source assumes the presence of a number of `node_modules`. 

https://github.com/jsonata-js/jsonata/issues/196 discusses a number of issues related to trying to get JSONata 1.5.3 to operate under Java 8.

xogeny:
I wonder if you could use rollup to address this?  I’ve been playing around with rollup in my TypeScript version. But nothing concrete yet.

michaelthoward:
Be advised that I have almost no exposure-to/experience-with the JavaScript/ES ecosystem. 

I googled and got high-level overviews of https://rollupjs.org and https://www.typescriptlang.org

It seems to me like the fundamental problem I am having with the babel-generated jsonata-es5 is the runtime dependency on a large set of node_modules in the *transpiled* source. 

I now understand TypeScript can also generate an ES5 version of the code. 
Q: What kind of runtime-library support is needed by a TypeScript-generated jsonata-es5 (or jsonata-es3) version? 
Q: Is it any *lighter-weight* than that generated by babel?

xogeny:
The point of `rollup` is to bundle all the JS (including dependencies) together into a single JS file.  This is commonly used when deploying web applications, but can be used to deploy any Javascript really.

I don't have much experience with `babel`, so I cannot make comparisons there.  But the point is that `rollup` gets read of `node_modules` completely.  The `node_modules` directory is part of the system for importing code as modules.  But `rollup` and other bundlers create bundles in which all used code is pulled out of bundles and inlined.  So you definitely shouldn't see any issues with needing files from `node_modules` when using a bundler.

But I don't really have any working solution for this in my version.  Just pointing out that it is a possibility.

andrew-coleman:
I don't have an immediate answer to this.  Adding 'help wanted' flag to open up to anyone with nashorn experience.

[...]

PR #231, opened 2018-05-25
Add a source directory

s100:
Per my comments under #214, #211 and #202, it looks as if having `package.json`'s `"module"` point at `jsonata.js` is causing things to break. This is because `jsonata.js` uses the new ES6 generator syntax such as `yield*`,  which many consumers (even ES6-aware consumers) do not understand.

To fix this, we need to split out the JSONata source to a separate directory `src`, and build our distributed ES6 module `./jsonata.js` from that source, eliminating the generator syntax as we do so.

This PR only does the first of these things, not the second, out of caution...

* Adds a `.editorconfig` file
* Adds trailing line breaks in several places
* Moves the source file `./jsonata.js` to `src/jsonata.js`
* Adjusts test files to point to the new location for the source
* Introduces an `npm` script, `"copy"`, which copies `src/jsonata.js` to `./jsonata.js` for distribution
* Adds `npm run copy` to the `"posttest"` script
* Adds the distributed file `./jsonata.js` to `.eslintignore` and `.gitignore`, but not `.npmignore`


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jsonata",
-  "version": "1.5.4",
+  "version": "1.5.5",
   "description": "JSON query and transformation language",
```
mattbaileyuk: Could you please not bump the version in this commit. We usually prefer to do the release separately from the change so we can bundle things up (Travis will automatically publish to npm on a new version number)

mattbaileyuk:
@s100 Thanks for the contribution, and I agree this looks like the right step forward towards solving the packaging problems highlighted in a number of issues.

Issue #158, opened 2018-02-19
JSONATA undefined in Internet explorer


hufedebe:
Hello everyone,

I just did a project heavily using JSONata, and it works perfectly well in Mozilla, Chrome, and Edge. But it doesnt work in Internet Explorer, it just shows the following error:

"SCRIPT5009: 'jsonata' is undefined", please help.

I'm just using it like this:

jsonata('string').evaluate(data);

andrew-coleman:
I suspect this is an ES6 issue.  Have you tried using the babel transpiled version `jsonata-es5.js`?

hufedebe:
Thank you Andrew, it almost works. In one part of the code i'm extracting properties normally with jsonata: info = jsonata('[xml.resume.{"date" : date, "field2": field2}]').evaluate(data); and then I want to add a counter, because its several objects, so I have this code: info = jsonata('$ ~> $map(function($v, $i) {$merge([{\'count\': $i+1}, $v])})').evaluate(info);. But that changes info into an Object Generator (according to Internet Explorer): [object Generator]. before I merge it with the count, it does output a correct object though, do I have to change the merge function? @andrew-coleman

andrew-coleman:
Does this same code evaluate correctly in other browsers?  I don't have Internet Explorer (or Windows) so there will be a lack of testing on my part here.

hufedebe:
Yes, in all other browsers it works fine, its having problems in Internet Explorer - it creates the [object Generator] instead of adding the "count" attribute to the existing object. Other than that, the other jsonata parts work well. its just the adding a "count" attribute to the existing object that works incorrectly, is there another way of obtaining the same result without using "$ ~> $map(function($v, $i) {$merge([{'count': $i+1}, $v])})"?

s100:
@hufedebe Can you provide minimal reproduction steps for this issue? At minimum please can you show us your HTML, what version of JSONata you are using, and what version of Internet Explorer you are using.

s100:
A minimal reproduction is:

```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="UTF-8">
        <title>JSONata test</title>
        <script src="jsonata-es5.js"></script>
        <script>
            const output = jsonata("$map([1..5], $string)").evaluate({});
            console.log(output)
        </script>
    </head>
    <body></body>
</html>
```

[...]

PR #235, opened 2018-06-11
Packaging fixes, phase two

s100:
Continues from #231. This one probably warrants a little debate.

Okay so the basic observation from #214, #211 and #202 is that several consumers are attempting to directly consume `./jsonata.js`, which currently makes use of the ES6 generator syntax, which does not have good support. The solution to this problem is to [compile the ES6 generator syntax away at build time using Babel](https://babeljs.io/docs/plugins/transform-regenerator), so that `./src/jsonata.js` can still use generators while the output file `./jsonata.js` does not. However, compiling the ES6 generator syntax away also requires us to pull in JSONata's first-ever third-party dependency, `regenerator-runtime`.

With this PR,

* The `npm` task `copy`, which simply copied `./src/jsonata.js` to `./jsonata.js`, is removed.
  * The development dependency on `ncp` is removed.
* A new `npm` task `mkdirp-es6` has been added, which creates a directory `./dist/es`, which is currently **not shipped or used** (but wait and see...).
  * A development dependency `mkdirp` has been added.
  * `./dist` has been added to `.eslintignore` and `.gitignore`.
* Another new `npm` task `babel-es6` has been added, which uses Babel and the `transform-regenerator` plugin to convert `./src/jsonata.js` (with ES6 generators) into `./dist/es/jsonata.js` (with no generators).
  * A development dependency on `babel-plugin-transform-regenerator` has been added.
* And a third new `npm` task `browserify-es6` has been added, which uses Browserify to prepare `./dist/es/jsonata.js` as the output UMD we are all familiar with, `./jsonata.js`.
* `./src/jsonata.js` now `require`s `regenerator-runtime`.
  * `regenerator-runtime` has been added to `package.json` as a dependency.

Basically: instead of just copying `./src/jsonata.js` to its destination at `./jsonata.js`, we now Babelify and then Browserify it.

This *should* resolve these consumers' problems...

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,6 @@
 coverage/
 node_modules/
+dist/
 doc/
 jsonata.js
 jsonata-es5.js
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 coverage/
+dist/
 doc/
 jsonata.js
 jsonata.min.js
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   ],
   "devDependencies": {
     "babel-cli": "^6.24.1",
+    "babel-plugin-transform-regenerator": "^6.26.0",
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "^1.5.2",
     "babel-runtime": "^6.23.0",
--- a/package.json
+++ b/package.json
@@ -14,8 +14,7 @@
     "pretest": "npm run lint",
     "mocha": "node ./node_modules/istanbul/lib/cli.js cover --report cobertura --report html ./node_modules/mocha/bin/_mocha -- \"test/**/*.js\"",
     "test": "npm run mocha",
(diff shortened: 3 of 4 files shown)


s100:
Looks like the tail recursion unit test failed on Node 6 but not on any other platform? I don't understand why...

s100:
I've ascertained that on Windows the new built `./jsonata.js` works fine in Chrome, Firefox and Edge. While it doesn't work in IE11 (as is to be expected due to the use of ES6 syntax like arrow functions), `./jsonata-es5.js` still works fine, as it did before.

andrew-coleman:
Thanks.  I think you are saying that this fix doesn't break anything.  However, have you ascertained whether it fixes any of the issues #214, #211, #202 & #158 ?  In particular #158 claims there is a problem in IE11.

s100:
This PR does not fix #158. This issue is not related to packaging concerns. I have opened a separate PR #243 which fixes #158.

s100:
However, I am reasonably sure that this will fix #211 and #214, and possibly #202.

PR #243, opened 2018-06-25
Better "is iterable" checking

### LINKED ISSUE #158: JSONATA undefined in Internet explorer

s100:
Fixes #158.

* Rename the method `isGenerator` to `isIterable` because that's what it does. I should have fixed this during #56 but never mind.
* Stop checking for function-ness. The line of code `Symbol.iterator in arg` compiles just fine in Babel, to something which uses `is-iterable` from `core-js`, and which works fine in Chrome/IE11 alike. However, the next line `typeof arg[Symbol.iterator] === 'function'` passes on Chrome but fails on IE11 - and is redundant, so the failure causes disagreeing results.


mattbaileyuk:
Closed and reopened to kick coveralls into life, but that didn't seem to happen. As coverage is fine in Travis runs I'm going to merge anyway.

PR #263, opened 2018-10-23
Handle DOMExceptions

### LINKED ISSUE #261: `$base64decode` failure gives an incorrect error message
If I run

```js
jsonata('$base64decode("not base64")').evaluate()
```

then a seemingly unrelated exception is thrown:

```
Uncaught TypeError: Cannot assign to read only property 'message' of object ''
```

This happens because internally the `$base64decode` method uses [`atob`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/atob). When `atob` fails, it throws a [`DOMException`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException). Later, `jsonata` attempts to reassign the `message` property of the `DOMException`:

```js
} catch (err) {
    // insert error message into structure
    err.message = lookupMessage(err);
    throw err;
}
```

However, while this is allowed for most `Error` objects, [the `message` property of a `DOMException` is read-only](https://developer.mozilla.org/en-US/docs/Web/API/DOMException/message). Attempting to assign to it causes the unrelated `TypeError` to be thrown instead, which is what we see output.

To solve this specific case, we could modify the `$base64decode` function to catch the `DOMException` and throw a good, honest `Error` instead. But to solve the general case, we could consider finding all the places where thrown exceptions are being mutated, and altering the code to create a new `Error` and throw that instead. Something like:

```js
} catch (err) {
  throw Error(lookupMessage(err))
}
```

Unfortunately this approach trashes the stack trace... so, perhaps some thought is required... hmm...

s100:
Fixes #261.

I fixed this by modifying (simplifying) `lookupMessage`.

* `lookupMessage` is now named `populateMessage`. Rather than return a string, it now populates `err.message` as a side-effect (or leaves it as is, if appropriate).
* Generally speaking, `populateMessage` now explicitly expects `err` to be either
  1. a plain old JavaScript object with an error `code`, thrown by `jsonata` itself, whose `message` should be populated using the appropriate template, or
  2. an actual `Error` object or similar (such as a `DOMException`) with no `code`, but with an existing `message` property which should not be touched.

Note that logic which automatically reified an error message of `"Unknown error"` was untested and therefore apparently unused, and so has been removed. The only way to hit this case was for a user to deliberately define a function which threw a non-`Error`, which I think is a rare case and not worth adding a test for.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
     "coveralls": "^3.0.0",
+    "domexception": "^1.0.1",
     "eslint": "^4.0.0",
     "eslint-plugin-ideal": "^0.1.3",
     "istanbul": "^0.4.5",
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
     "coveralls": "^3.0.0",
-    "domexception": "^1.0.1",
     "eslint": "^4.0.0",
     "eslint-plugin-ideal": "^0.1.3",
     "istanbul": "^0.4.5",
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -4939,18 +4939,14 @@ var jsonata = (function() {
 
     /**
      * lookup a message template from the catalog and substitute the inserts.
-     * Populates `err.message` with the substituted message.
+     * Populates `err.message` with the substituted message. Leaves `err.message`
+     * untouched if code lookup fails.
      * @param {string} err - error code to lookup
      * @returns {undefined} - `err` is modified in place
      */
     function populateMessage(err) {
         var template = errorCodes[err.code];
-        if(typeof template === 'undefined') {
-            if(typeof err.message === 'undefined') {
-                err.message = 'Unknown error';
-            }
-            // Otherwise retain the original `err.message`
-        } else {
+        if(typeof template !== 'undefined') {
             // if there are any handlebars, replace them with the field references
             // triple braces - replace with value
             // double braces - replace with json stringified value
@@ -4962,6 +4958,7 @@ var jsonata = (function() {
             });
             err.message = message;
         }
+        // Otherwise retain the original `err.message`
     }
 
     /**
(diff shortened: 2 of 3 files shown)


s100:
Note that unfortunately it's not possible to directly test the behaviour of `atob` since it's not available in our testing environment.

s100:
Ah, well, it seems that the `domexception` module doesn't work on Node.js v4? However, Node.js v4 reached EOL in [April 2018](https://github.com/nodejs/Release#end-of-life-releases) so a case can be made that support for it could be dropped... I will leave this call up to the reviewer...

mattbaileyuk:
`node-red` is one of the bigger consumers of `jsonata`, and I believe that still supports node v4, so I don't think we have a choice yet whether to drop support. Having said that, I think it consumes `jsonata-es5.js` so it might still be safe...? However, that would put a risk that `jsonata` doesn't work in it and we wouldn't know until they tried to use it (they fix the version, so wouldn't pick this up right away).

mattbaileyuk:
Is it worth adding a test case to https://github.com/jsonata-js/jsonata/tree/master/test/test-suite/groups/encoding that would catch the original issue in https://github.com/jsonata-js/jsonata/issues/261 so we don't regress?

s100:
@mattbaileyuk You'll notice we already have a test case for `$base64decode` decoding failure, case 003, which unfortunately does not exercise the bug.

`jsonata` uses a different mechanism for Base64 decoding depending on whether it is running in a browser or not. In a browser it uses `window.atob` but in our tests it uses Node.js's `Buffer`. The existing tests for `$base64decode` all use `Buffer` - `window.atob` and `window` itself do not exist with our current testing arrangements. Notice that this case is also excluded from our code coverage metrics:

```js
        var atob = typeof window !== 'undefined' ?
            /* istanbul ignore next */ window.atob :
            function(str) {
                // Simply doing `new Buffer` at this point causes Browserify to pull
                // in the entire Buffer browser library, which is large and unnecessary.
                // Using `global.Buffer` defeats this.
                return new global.Buffer(str, 'base64').toString('binary');
            };
```

mattbaileyuk:
Ah, yes, good spot. Something to think about later.

Thanks for the PR!

PR #266, opened 2018-10-30
consistent handling of function params for HOFs

### LINKED ISSUE #233: Inconsistency between $map and $filter/$sift
The `$map` and `$filter` functions have similar definitions. They both take an array of values and a function with the following signature (from the docs):

```
function(value [, index [, array]])
```

This suggests that `$map` and `$filter` can handle functions that take one, two, or three arguments. But only `$map` seems to work like this. `$filter` chokes if you give it a function that can't take all three arguments.

Let's use the `$boolean` function as an example. It only takes one argument, yet this map expression works as you'd expect:

```
$map([0, 1, 2], $boolean)  // returns [false, true, true]
```

Whereas using `$boolean` in a filter expression causes an error:

```
$filter([0, 1, 2], $boolean)  // returns Argument 2 of function "filter"
                                     does not match function signature
```

You can make this work by wrapping the call to `$boolean` in a lambda function:

```
$filter([0, 1, 2], function($x){$boolean($x)})  // returns [1, 2]
```

It might appear that this lambda function only takes one argument, but lambdas without type signatures will happily accept any number of arguments (try it!). Add a type signature specifying a single argument, and the expression fails again: 

```
$filter([0, 1, 2], function($x)<x:b>{$boolean($x)})  // fails!
```

If you play around with the signature it's clear that any function incompatible with three arguments causes an error:

```
$filter([0, 1, 2], function($x)<xx:b>{$boolean($x)})  // fails!

$filter([0, 1, 2], function($x)<xxx:b>{$boolean($x)})  // returns [1, 2]

$filter([0, 1, 2], function($x)<x+:b>{$boolean($x)})  // returns [1, 2]

$filter([0, 1, 2], function($x)<xxxx:b>{$boolean($x)})  // fails!
```

This looks like a bug in the `$filter` function. While `$map` checks how many arguments your function takes and only passes that number, `$filter` always passes all three values. This happens to work for lambdas but it can cause issues with built-in or third-party functions.

`$sift`, the equivalent of `$filter` but for objects, has the same problem:

```
$sift({"a": 1}, $boolean)  // fails!
```

### LINKED ISSUE #250: Should $each accept functions that take a single argument?
The following expression returns an error:

```
$each({"hello": "world"}, $uppercase)  // Error: Argument 2 of function "each"
                                             does not match function signature
```

This makes sense. According to the docs, the second parameter of `$each` is a function that takes two arguments, `value` and `name`. And `$uppercase` only takes one argument.

However the docs also say that `name` is optional, which suggests (to me at least) that `$each` should work with functions that take a single argument. Other higher order functions like `$sift`, `$map` and `$filter` either already work like this or are being updated to do so (see [this issue](https://github.com/jsonata-js/jsonata/issues/233)). Is there any reason why `$each` should not follow suit?

andrew-coleman:
The handling of optional arguments for the function arg passed to $filter, $each & $sift should be consistent with the correct behaviour in $map.
resolves #233 
resolves #250

### DIFF
--- a/test/implementation-tests.js
+++ b/test/implementation-tests.js
@@ -384,7 +384,7 @@ describe("Tests that bind Javascript functions", () => {
         });
     });
 
-    describe("sift with a user-defined Javascript function", function() {
+    describe("$sift with a user-defined Javascript function", function() {
         it("should return result object", function() {
             var expr = jsonata("$sift({'one': 1, 'four': 4, 'nine': 9, 'sixteen': 16}, $even)");
             expr.assign("even", function(num) {
--- a/test/implementation-tests.js
+++ b/test/implementation-tests.js
@@ -372,6 +372,42 @@ describe("Tests that bind Javascript functions", () => {
         });
     });
 
+    describe("filter with a user-defined Javascript function", function() {
+        it("should return result object", function() {
+            var expr = jsonata("$filter([1,4,9,16], $even)");
+            expr.assign("even", function(num) {
+                return num % 2 === 0;
+            });
+            var result = expr.evaluate(testdata2);
+            var expected = [4, 16];
+            expect(result).to.deep.equal(expected);
+        });
+    });
+
+    describe("sift with a user-defined Javascript function", function() {
+        it("should return result object", function() {
+            var expr = jsonata("$sift({'one': 1, 'four': 4, 'nine': 9, 'sixteen': 16}, $even)");
+            expr.assign("even", function(num) {
+                return num % 2 === 0;
+            });
+            var result = expr.evaluate(testdata2);
+            var expected = {'four': 4, 'sixteen': 16};
+            expect(result).to.deep.equal(expected);
+        });
+    });
+
+    describe("$each with a user-defined Javascript function", function() {
+        it("should return result object", function() {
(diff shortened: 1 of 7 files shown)


test/implementation-tests.js
```diff
@@ -372,6 +372,42 @@ describe("Tests that bind Javascript functions", () => {
...
+            expr.assign("even", function(num) {
+                return num % 2 === 0;
+            });
+            var result = expr.evaluate(testdata2);
+            var expected = [4, 16];
+            expect(result).to.deep.equal(expected);
+        });
+    });
+
+    describe("sift with a user-defined Javascript function", function() {
```
mattbaileyuk:
```suggestion
    describe("$sift with a user-defined Javascript function", function() {
```

test/implementation-tests.js
```diff
@@ -372,6 +372,42 @@ describe("Tests that bind Javascript functions", () => {
         });
     });
 
+    describe("filter with a user-defined Javascript function", function() {
```
mattbaileyuk:
```suggestion
    describe("$filter with a user-defined Javascript function", function() {
```

PR #270, opened 2018-11-01
Update dependencies

mattbaileyuk:
Updating the `eslint` and `regenerator-runtime` dependencies to their latest versions. Neither require any code changes to be made, so this is just a simple update of `package.json` to keep up with current.


mattbaileyuk:
Looks like `eslint` does not work on nodeJS `4.x` as of `eslint@5.x`, so we'll have to leave that back-versioned until `jsonata` drops support for that.

Issue #247, opened 2018-07-11
Error selecting or performing operations on JSON objects containing "sequence" key


apita-carecloud:
node v8.9.3
jsonata v1.5.4
Also able to duplicate on [JSONata Exerciser](http://try.jsonata.org/)

Attempt any selection on an object with a "selection" key.

### The following example works a s expected:
**JSON**
`{
  "foo": {
    "sequence1": 123,
    "bar": "baz"
  }
}`

**JSONata**
`foo.bar`

**Result**
`"baz"`

### When the key is changed to "sequence", it fails:
**JSON**
`{
  "foo": {
    "sequence": 123,
    "bar": "baz"
  }
}`

**JSONata**
`foo.bar`

**Result**
`n.value is not a function`

Repo search revealed 'sequence' being used in https://github.com/jsonata-js/jsonata/blob/7aa65f2295e0605a284e1f1083cf528aceb4ba79/src/jsonata.js#L1557-L1566

PR #292, opened 2019-01-03
Fix performance regression

andrew-coleman:
The fix for #247 introduced a significant performance regression.
This PR rewrites that bit of code to recover the loss.

### DIFF
--- a/test/implementation-tests.js
+++ b/test/implementation-tests.js
@@ -140,7 +140,7 @@ describe("Functions with side-effects", () => {
 
     describe("$now() returns timestamp with defined format", function() {
         it("should return result object", function() {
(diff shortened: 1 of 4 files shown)


test/implementation-tests.js
```diff
@@ -140,7 +140,7 @@ describe("Functions with side-effects", () => {
 
     describe("$now() returns timestamp with defined format", function() {
         it("should return result object", function() {
-            var expr = jsonata("$now('[h]:[M][P] [z]')");
+            var expr = jsonata("$now('[h]:[M01][P] [z]')");
```
mattbaileyuk: I believe this is unrelated to the rest of the fixes, but agree with the change so not too worried that it's here.

PR #293, opened 2019-01-10
$each should not result in undefined values in the result object

andrew-coleman:
`$each()` returns an array of values by applying a supplied function to each name/value pair in a given object.  If that function returns `undefined` (no match / empty sequence), then that `undefined` value is getting added to the result sequence.  It shouldn't - it should be flattened out.


mattbaileyuk:
@andrew-coleman Good spot; planning to push out 1.6.4 with this?

Issue #349, opened 2019-08-20
Current Unit Tests Fail When run in a non GMT+0 Environment


jhorbulyk:
```
  2 failing

  1) JSONata Test Suite - async mode
       Group: function-tomillis
         case004.json: $toMillis("2017-10-30T15:28:34"):

      AssertionError: expected 1509373714000 to deeply equal 1509377314000
      + expected - actual

      -1509373714000
      +1509377314000
      
      at getBasePromise.then.then.newArgs (node_modules/chai-as-promised/lib/chai-as-promised.js:302:22)
      at <anonymous>

  2) JSONata Test Suite
       Group: function-tomillis
         case004.json: $toMillis("2017-10-30T15:28:34"):

      AssertionError: expected 1509373714000 to deeply equal 1509377314000
      + expected - actual

      -1509373714000
      +1509377314000
      
      at Context.<anonymous> (test/run-test-suite.js:120:52)
```
This failure is because the test suite expects the expression to evaluate `$toMillis("2017-10-30T15:28:34")` as a GMT timestamp while my machine reads it as a local timestamp (in GMT+01:00).  This can be solved by making the test non-ambiguous with `$toMillis("2017-10-30T15:28:34Z")`.

mattbaileyuk:
Fixed in #350 

Thanks for spotting *and* fixing this 👍

PR #350, opened 2019-08-20
Fix error with ambiguous timezone in tests.

jhorbulyk:
Described in #349.


mattbaileyuk:
@jhorbulyk Good spot, and thanks for the fix!

Issue #103, opened 2017-12-07
Proposal: URL/URI Helper Functions


jhorbulyk:
It would be useful if the following JavaScript string operators were exposed to JSONata:
* [`encodeURIComponent()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)
* [`encodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI)
* [`decodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI)
* [`decodeURIComponent()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent)

This functionality would be useful in order to write transformations which produce or consume URLs/URIs.  Would adding such functionality be in the spirit of the library's intended purposes?

xogeny:
Can't you simply add these via `assign`/`bindings`/`registerFunction`?

jhorbulyk:
@xogeny You could register these functions however they encompase pretty standard functionality that one would want in most places which is why adding them to the set of built-in functions would be pretty desireable IMO.

xogeny:
Do you think this could be addressed by having a plugin system for registering functions in bulk?  That way, people could pick and choose what "libraries" of functions they wanted to have pre-defined in their application *and* it could even provide a mechanism to provide such libraries as packages in `npm` for easy use.

jhorbulyk:
A plugin system could be a useful JSONata feature.  My thoughts:

- It isn't necessary for this particular request
- These exist as native functions in the underlying language (JavaScript).  It seems sensible that they could also exist as native functions in JSONata.
- The cost of  overhead of a plugin system may not be worth it given the relative symplicity of JSONata.
- JSONata appears useful as transformation language.  I'm less certain about JSONata's utility as a framework for transformations.
- I can't think of use cases in which a user wouldn't want functions such as `encodeURIComponent()` to be available.

xogeny:
All good comments.  Regarding your last point, the issue is really just making sure the global namespace doesn't get too polluted.  One could say "I can't think of a use cases in which a user wouldn't want ... mongodb/fetch/node/DOM/whatever ...functions to be available".  But you make a good point that these are native to Javascript and don't introduce dependencies.

andrew-coleman:
[...]

PR #357, opened 2019-08-26
Add URL/URI Helper Functions

jhorbulyk:
As discussed in #103


jhorbulyk:
@andrew-coleman Do you have any preference for one single jsonata error code or should I add an error code for each of the 4 functions?

andrew-coleman:
One error code is fine.  You could add the function name as an insert into that error.  Thanks.

jhorbulyk:
@andrew-coleman Updated.

PR #360, opened 2019-09-02
Support nodeJS 12.x

mattbaileyuk:
nodeJS version 12.x is the current latest version and will become the new LTS in October, so we should get Travis testing against this.

As a non-LTS odd numbered version, we can get rid of testing against v11.


mattbaileyuk:
There's no code changes here, so it looks like a change in nodeJS 12 is breaking a couple of the tests 😞

jhorbulyk:
> There's no code changes here, so it looks like a change in nodeJS 12 is breaking a couple of the tests

And with the my new encodeURL functions too ....

andrew-coleman:
@jhorbulyk I wonder if you could look at these two failing test cases, please?  I'm not keen to publish a new version on npm that doesn't build cleanly on node v12 (as well as 8 & 10).  Many thanks.

mattbaileyuk:
I've so far tracked it down to https://github.com/jsonata-js/jsonata/blob/master/src/jsonata.js#L1841 (that is, that block) returning a different result between nodeJS 10 and 12 (suggesting that stringify has changed behaviour slightly for unicode characters?). I'm unsure at the moment how to make that work on both, other than changing the test case to use a different value.

mattbaileyuk:
It's not a bug in nodeJS 12, it is in fact an intentional change (https://github.com/tc39/proposal-well-formed-stringify) which made it into ES2019 (and thus into the latest versions of nodeJS).

@jhorbulyk Do you think there is an alternative value we can use in the test case that will give the same result on nodeJS 12 and earlier versions? The problem I've had is that it the only way I can find to cause `encodeURI()` to generate the error you were testing for is to use unicode values which now differ with `JSON.stringify()` between versions.

mattbaileyuk:
@andrew-coleman That seems like a good way to go with the test change 👍

PR #362, opened 2019-09-04
Enhance $reduce to support 4 arg function

### LINKED ISSUE #102: Need a good way to locate elements in an array, or at least that element's index
I'm trying to do something that I thought would be easy, but running into some roadblocks...

As an example, let's try to convert an input Month abbreviation into the Date month number (0-based). It's easy to find out if the incoming string is in the array of months, using `payload in $months`, but finding out what position it occupies in the array is not so easy:
```
(
    $months := [
        "Jan", "Feb", "Mar", "Apr", "May", "Jun",
        "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"
    ];
    $months~>$map(function($val, $idx) {
        $val = payload ? $idx : undefined
    })
)
```
A bit clunky, perhaps, but it returns the index number for every array element that is equal to the incoming string. I tried using the `$reduce($array, function($total, $value))` -- but unlike the Javascript function of the same name, there is no index value passed as the 3rd argument. **At a minimum, I think the `$reduce()` function should be extended to also pass $index as an optional argument**.

But perhaps a pair of new functions would be more intuitive and simpler to use, such as:
* `$find(array, input)` and the overloaded `$find(array, function($value))`
* `$indexOf(array, input)` and the overloaded `$indexOf(array, function($value))`

Both functions take an array of "things", and either a similar "thing" to find, or a function that determines which element is a match (returns true or false).

andrew-coleman:
The second argument can now be an arity-4 function with the following signature
`myfunc($accumulator, $value[, $index[, $array]])`

resolves #102

### DIFF
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -1787,7 +1787,7 @@ var jsonata = (function() {
         "D3020": "Third argument of split function must evaluate to a positive number",
         "D3030": "Unable to cast value to a number: {{value}}",
         "D3040": "Third argument of match function must evaluate to a positive number",
-        "D3050": "First argument of reduce function must be a function with two arguments",
+        "D3050": "The second argument of reduce function must be a function with at least two arguments",
         "D3060": "The sqrt function cannot be applied to a negative number: {{value}}",
         "D3061": "The power function has resulted in a value that cannot be represented as a JSON number: base={{value}}, exponent={{exp}}",
         "D3070": "The single argument form of the sort function can only be applied to an array of strings or an array of numbers.  Use the second argument to specify a comparison function",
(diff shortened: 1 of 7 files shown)


src/jsonata.js
```diff
@@ -1787,7 +1787,7 @@ var jsonata = (function() {
         "D3020": "Third argument of split function must evaluate to a positive number",
         "D3030": "Unable to cast value to a number: {{value}}",
         "D3040": "Third argument of match function must evaluate to a positive number",
-        "D3050": "First argument of reduce function must be a function with two arguments",
+        "D3050": "The second argument of reduce function must be a function with at least two arguments",
```
mattbaileyuk: Was curious about this change then realised that this was already wrong, and it should have said the second argument anyway.

Issue #332, opened 2019-05-17
Blocks does not work with nested array structure


yahelnachum:
# Summary
In the [jsonata playground](http://try.jsonata.org/) I was trying out [parenthesized expressions and blocks](http://docs.jsonata.org/composition#parenthesized-expressions-and-blocks) and came across some inconsistent output. I used blocks to turn a selection into an array so that I could pick a single index. Then when I tried adding a key to step into that single index it seemed to revert back to a selection.

# More Details

JSON Input 1:
```
[
  {
    "phone": [
      {"number" : 0}
    ]
  },
  {
    "phone": [
      {"number" : 1}
    ]
  },
  {
    "phone": [
      {"number" : 2}    
    ]
  }
]
```

Then I tried a series of JSONata expressions

----
Expression 1:
```
phone
```

Actual Output 1:
```
[
  {"number":0},
  {"number":1},
  {"number":2}
]
```

---
Expression 2:
```
(phone)[0]
```

Actual Output 2:
```
{
  "number": 0
}
```
----
Expression 3:
```
(phone)[0].number
```

Actual Output 3:
```
[
  0,
  1,
  2
]
```

----
I would have expected the output of `Expression 3` to be the following given that I was able to select a single index in `Expression 2`.

Expected Output 3:
```
0
```

This seems to me like a bug in the code somewhere, but if there is an explanation for this behavior then please let me know what that is. Thanks in advance for reading!

yahelnachum:
My co-worker was trying some other inputs and got the expected behavior
JSON input 2:
```
{
  "account": [
    {
      "phone": [
        {"number" : 0}
      ]
    },
    {
      "phone": [
        {"number" : 1}
      ]
    },
    {
      "phone": [
        {"number" : 2}    
      ]
    }
  ]
}
```

----
Expression 4:
```
(account.phone)[0].number
```

Actual Output 4:
```
0
```

----

So for some reason wrapping my `JSON input 1` in a JSON object gives me the correct behavior.

yahelnachum:
Just tried another input with a nested array with the same `Expression 3` and it worked this time. Still not sure why its behaving differently.

JSON input 3:
```
[
  [
    {
      "phone": [
        {"number" : 0}
      ]
    },
    {
      "phone": [
        {"number" : 1}
      ]
    },
    {
      "phone": [
        {"number" : 2}    
      ]
    }
  ]
]
```

----
Expression 5:
```
(phone)[0].number
```

Actual Output 5:
[...]

Issue #316, opened 2019-03-07
Empty Array input always renders empty object output no matter what the query is


apoggi-carecloud:
Input:
`[]`
will always result in 
Output:
`{}`
Example query:
`{"hello": "world"}`

mattbaileyuk:
@apoggi-carecloud Would http://try.jsonata.org/SycA90CIE describe your situation?
That looks like a bug to me, whereby the query which is being used to build a new output (and thus shouldn't care about the input) is being ignored in favour of processing of the input empty array. (Applying a function does work, so it's not getting `{}` is _every_ case)

apoggi-carecloud:
Your linked example is exactly the issue I am having.

andrew-coleman:
This issue is related to how JSONata handles input arrays.  There is a lengthy discussion in #170 on this topic.  I'm currently prototyping a solution that uses explicit syntax to determine whether to map over the input array verses just use it as a single context value.

andrew-coleman:
resolved by #363

Issue #170, opened 2018-03-13
Issues with path processing


xogeny:
Another issue I see with evaluation semantics.  Consider the following expression (input doesn't matter): `[1, 2, 3].$`.  The result of evaluating this is `[1, 2, 3]`.  The idea here is that because `.` is a map operation, we are simply mapping over each element in `[1, 2, 3]` and then for each successive value (*i.e.,* `1`, `2`, and `3`) we are evaluating the expression `$`.

Now this construction is useful because we see that adding `.$` to any expression that evaluates to an array should have no impact.  In a sense, this is (or at least one would certainly expect) that this would be an identity transformation in the same what that surrounding an atomic expression with a block (*i.e.,* `(...)`) shouldn't change the value (assuming the block doesn't impact the precedence of operators, of course).

The issue I see in `v1.5.x` is that it there are special cases in the evaluation code and these special cases generally lead to surprising results because they lead to inconsistencies in evaluation.  The following is an example...

Consider we evaluate the expression `[1, 2, 3]`.  Of course, our result will be `[1, 2, 3]`.  In the same way, if we evaluate:

```
($v := [1,2,3]; $v)
```

...we also get `[1, 2, 3]`.  This is no surprise because we have just effectively wrapped a block around our original expression.  But the problem is with these special cases.  In particular, the current evaluation code actually handles the first step in a path in a special way.

Conceptually speaking, the evaluation code needs to determine if the first step in the path is effectively independent of the input value (*i.e.,* evaluates to a constant).  This is important because (as far as I can tell), if the path *does* depend on the input, then there is an implicit map operation applied by the first expression (*i.e.,*, you could conceptually speaking prepend a `$.` to the start of your expression).  But if it *doesn't* depend on the input, then the first step isn't an expression we map over but instead **amounts to the effective input for the remainder of the path**.

There are two issues here.  First, these semantics are inconsistent (i.e., the role of the first step changes depending on context in complex and hard to describe ways).  But the bigger problem, in some sense, is the fact that it doesn't actually handle all cases correctly.

To demonstrate this, consider the following table:

|  Expression | Result |
| ----------- | ------ |
| `[1, 2, 3]` | `[1, 2, 3]` |
| `[1, 2, 3].$` | `[1, 2, 3]` |
| `($v := [1,2,3]; $v)` | `[1, 2, 3]` |
| `($v := [1,2,3]; $v).$` | `[1, 2, 3]` |

[...]

PR #363, opened 2019-09-05
input arrays treated as single value unless explicitly mapped using $.

andrew-coleman:
This is a proposed solution to bugs #332 and #316 which I think also addresses the comments that were discussed in #170 

This change only affects how top level JSON arrays are handled in the input data.  If you wish to explicitly map the JSONata expression to _each_ item in the input array, then the expression will need to explicitly declare that by starting it with `$.`  Otherwise the input array as a whole will be treated as the single context value of the expression.

I would argue that this is not a breaking change because the current behaviour is somewhat broken (hence the bug reports).  Although I have had to modify a couple of test results (edge cases, I think).


src/jsonata.js
```diff
@@ -1904,6 +1904,12 @@ var jsonata = (function() {
                 timestamp = new Date();
                 exec_env.timestamp = timestamp;
 
+                // if the input is a JSON array, then wrap it in a singleton sequence si it gets treated as a single input
```
mattbaileyuk: Minor typo: `si` -> `so`

PR #371, opened 2019-10-01
Support for data joins and positional variables

### LINKED ISSUE #187: $indexOf function for array enhancement request
I wanted to leverage the location of elements in an array to do something, but couldn't figure out how to get the index of a particular array element matching search criteria.  I would like to see something like:
```
$indexOf(Account.Order[OrderID="order103"])
or
Account.Order.indexOf(OrderID="order103")
or
Account.Order[OrderID="order103"].index
```
Above are listed in order of preference but would be happy with any of them.
Also posted here if you'd prefer I just use a function:
https://stackoverflow.com/questions/49655474/how-to-get-index-of-element-in-array-in-jsonata

### LINKED ISSUE #333: how to "join" objects based on a common key
Apologies I'm new to JSONata hopefully a dumb question and I just saw there is a slightly similar but not identical question to mine here:
https://github.com/jsonata-js/jsonata/issues/275

Our use case is we will have JSON with arrays from different REST apis whose entries we will need to match based on a common key and then transform them together into a single array of combined objects (although not just merge them actually pick certain properties or otherwise transform the property values so slightly different than #275 above which only merged the objects).

(also our goal in using JSONata is to have something a little friendly for non/semi-technical users and I'm hoping maybe there's a simpler solution than #275 which involves using a function we were hoping our use of JSONata would avoid that)

The test example I created in the JSONata Exercisor is as follows:

```json
{
	"Employee": [{
			"SSN": "496913021",
			"FirstName": "Fred",
			"Surname": "Smith"
		},
		{
			"SSN": "496737199",
			"FirstName": "Darren",
			"Surname": "Cruse"
		}
	],
	"Contact": [{
			"ssn": "496913021",
			"Phone": [{
					"type": "home",
					"number": "0203 544 1234"
				},
				{
					"type": "office",
					"number": "01962 001234"
				},
				{
					"type": "mobile",
					"number": "077 7700 1234"
				}
			]
		},
		{
			"ssn": "496737199",
			"Phone": [{
					"type": "home",
					"number": "3146458343"
				},
				{
					"type": "mobile",
					"number": "315 782 9279"
				}
			]
		}
	]
}
```

So I was trying to match the array elements by the social security number ("SSN" in Employee and "ssn" in Contact).

This I thought I was getting close with but could never get it to work right:
```
$.Employee.{
    "SSN": $.SSN,
    "name": FirstName & " " & Surname,
    "phone": $$.Contact[ssn=$.SSN].Phone[type="mobile"].number	
}
```

But I can't seem to get the Contact to match by the Employee SSN I get:
```json
[
  {
    "SSN": "496913021",
    "name": "Fred Smith"
  },
  {
    "SSN": "496737199",
    "name": "Darren Cruse"
  }
]
```

UPDATE:  Just played a little more looking at the #275 solution and I did get this to work:
```
Employee.{
    "SSN": SSN,
    "name": FirstName & " " & Surname,
    "phone": ($$.Contact ~> $filter(function($contact){$contact.ssn = SSN})).Phone[type="mobile"].number	
}
```

Still wondering is this the simplest it can be?

If so then I'm wondering could I abstract it away somehow?  e.g. Could I hide that filter behind a custom function I would provide something like:

```
Employee.{
    "SSN": SSN,
    "name": FirstName & " " & Surname,
    "phone": $matchKey($$.Contact, "ssn", SSN).Phone[type="mobile"].number	
}
```
should that be possible?

andrew-coleman:
This PR introduces two new operators that can be used within a path expression:
- `@$var` is used on a map step (`.`) to bind a variable to the current context item.
- `#$var` is used to bind a variable to the current context position (in the sequence).

The former can be used to perform data cross referencing and joins - resolves #333.
The latter resolves #187 
The documentation has been updated with details of these operators, plus a section on the explaining path processing model.


test/parser-recovery.js
```diff
@@ -87,14 +91,17 @@ describe('Invoke parser with incomplete expression', function() {
...
+                        "stages": [
                             {
+                                "expr": {
                                 "type": "error",
                                 "error": {
                                     "code": "S0207",
                                     "position": 8,
                                     "token": "(end)"
-                                }
+                                }                                },
```
mattbaileyuk: Some strange formatting going on here; surprised the linter didn't pick this up.

test/parser-recovery.js
```diff
@@ -288,11 +307,6 @@ describe('Invoke parser with incomplete expression', function() {
             var expr = jsonata('$equals3lucy[0].UnstructuredAnswers^()[0].Text', { recover: true });
```
mattbaileyuk: I can see where this expression has come from! While editing around here it might be good to remove the reference to the third-party company.

test/parser-recovery.js
andrew-coleman: fixed

test/parser-recovery.js
andrew-coleman: fixed

jugaadi:
Can't we reduce it to just `@var`?
```
library.loans@l.books@b[$l.isbn=$b.isbn].{
  'title': $b.title,
  'customer': $l.customer
}
```
Same with `#var`
```
library.books#i['Kernighan' in authors].{
  'title': title,
  'index': $i
}
```

Alternative : An underscore to represent *context* computation. It can be extended just like functions.
Syntax        : `_<keyword specific symbol><var>`, `_<keywords>$<var>`, or `_<keyword>` for current context.
Context Bindings :
- Index - `_@i` or `_index$i` or `$_index`
- Instance - `_#o` or `_instance$o` or `$_instance`
- Type - `_%t` or `_type$t` or `$_type`
- Parent - `_$` or `_parent$p` or `$_parent`

Example     : 
```
library.books_#i['Kernighan' in authors].{
  'title': title,
  'index': $i
}
```
```
library.loans_@l.books_@b[$l.isbn=$b.isbn].{
  'sno': $_index
  'title': $b.title,
  'customer': $l.customer
}
```

andrew-coleman:
Thanks for the feedback @jugaadi.  In one of my early prototypes, I did experiment with dropping the '$'.  However, I felt that the dollar made the variable token consistent across the language.  It is also used when binding variables using the `:=` operator, even though it could be argued that that is not necessary either.

Issue #369, opened 2019-09-30
Proposal: Add $assert() function


jhorbulyk:
Originally proposed in https://github.com/jsonata-js/jsonata/issues/167

Once JSONata has an `$error()` function, the most common use case becomes building assert statements.

Syntax:
`$assert(condition, message)`
If `condition` is true, the function returns undefined.  If the condition is false, an exception is thrown with the message as the message of the exception.

andrew-coleman:
Resolved by #375

PR #375, opened 2019-10-02
Add $assert() function.

jhorbulyk:
Function described in https://github.com/jsonata-js/jsonata/issues/369


src/functions.js
```diff
@@ -1822,6 +1822,25 @@ const functions = (() => {
...
 
+    /**
+     *
+     * @param {boolean} condition - the condition to evaluate
+     * @param {string} [message] - the message to attach to the error
+     * @throws custom error with code 'D3137'
+     * @returns {undefined}
+     */
+    function assert(condition, message) {
+        if(condition) {
```
andrew-coleman: isn't this the wrong way round?  it should throw if the condition is false

Issue #208, opened 2018-04-19
Request: function or functions for testing data type of a value


michaelthoward:
Per [this question on stackoverflow](https://stackoverflow.com/questions/49712517/jsonata-method-or-function-for-testing-json-value-datatype/49900869), JSONata needs a function functions for testing JSON data type. 

Seems this could be predicates for testing individual data types (e.g. isBoolean(), isNumber(), etc.) and/or a function which returned the data type as a string.

basel-shishani:
+1
I believe this feature is essential as it's very common to have structures with anyOf possible types.

jhorbulyk:
**Proposal for syntax:** 
`$type(value)`
Returns:
- (non-string) `undefined` if `value` is undefined
- (string) `'null'` if `value` is null
- (string) `'number'` if `value` is a number
- (string) `'string'` if `value` is a string
- (string) `'boolean'` if `value` is a boolean
- (string) `'array'` if `value` is an array
- (string) `'object'` if `value` is an object
- (string) `'function'` if `value` is a function

andrew-coleman:
Resolved by #381

PR #381, opened 2019-10-16
Add typeOf() function.

jhorbulyk:
Proposed in https://github.com/jsonata-js/jsonata/issues/208


src/functions.js
```diff
@@ -1841,6 +1841,39 @@ const functions = (() => {
...
+            return 'null';
+        }
+
+        if (isNumeric(value)) {
+            return 'number';
+        }
+
+        if(['string', 'boolean'].includes(typeof value)) {
+            return typeof value;
+        }
```
andrew-coleman: Array.includes() is an ES6 function, which means that the transpiled jsonata-es5.js will not work in ES5 environments such as IE unless we explicitly put the polyfill into [polyfill.js](https://github.com/jsonata-js/jsonata/blob/master/polyfill.js).  
However, in this case, I think it would be easier to remove this block and replace it with two `if` statements for the string and boolean cases.

docs/object-functions.md
```diff
@@ -53,12 +53,27 @@ __Examples__
...
 Deliberately throws an error with an optional `message`
 
- ## `$assert()`
+## `$assert()`
 __Signature:__`$assert(condition, message)`
 
 If condition is true, the function returns undefined. If the condition is false, an exception is thrown with the message as the message of the exception.
+
+## `$typeOf()`
+__Signature:__`$typeOf(value)`
```
andrew-coleman: I think I prefer the name `$type` as you proposed originally in #208.  In general I favour names as single words, provided the word is a meaningful representation of what the function does.  I think in this case `$type` would be ok?

jhorbulyk: Sure.  Most languages (e.g. JavaScript) call the operator `typeof` so I went with that pattern.  If single word function names are preferred, we can use that.

andrew-coleman:
excellent - many thanks!

PR #397, opened 2020-01-07
Update jsonata.d.ts

OpportunityLiu:
- Add an overload of `evaluate` to clarify async call & sync call.
- Add `this` parameter to `registerFunction`

Signed-off-by: Opportunity opportunity@live.in


andrew-coleman:
Many thanks!

PR #400, opened 2020-01-23
Parent

### LINKED ISSUE #299: Parent operator proposal
A while ago I said that I would describe the way a parent operator works in legacy programming languages.
Since then I discovered the power of [Symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol). 

In legacy we would work with classes so to add behaviour to data we would create an access class as a layer between the framework(jsonata) and the data. In there pointers to the data could be stored to indicate, for instance, parent of current data.

Drawback of this solution is that every call to the evaluation part has to be redirected to this class. That is quite a bit of interfacing.

Now with javascript we have symbols as a metadata annotation that can be added directly to the data without pollution the look and feel. We could add either a reference, a copy or a path so that we can easily retrieve the parent.
In the jsonata common scope add: 
```
var parentSymbol = Symbol('parent');
```
and for instance use a standard name in the `evaluateName` method:
```
            if(expr.value === "_parent" && input[parentSymbol]){
                result = input[parentSymbol];
            } else {
                result = input[expr.value];
            }
```
Then whenever a name is traversed, the symbol must be added, with the parent:
```
            if(typeof result === "object" && result !== null && !result[parentSymbol]){
                result[parentSymbol] = input;
            }
```
In the debugger we can see the symbol but it cannot be accessed outside of the jsonata scope.

This above example is not complete since it does not deal with arrays, but you get the point.
A similar approach can be used to store paths. This to do even more complicated accessing.

[Here is an idea of path + access class in java](https://github.com/jmespath/jmespath.js/issues/22)
Let me know what you think.
Cheers
Martin

andrew-coleman:
New syntax for accessing the parent object of the current context value.
Resolves #299


loganvolkers:
Added some more error test cases in #402 to confirm the theory that `%%` could be used for generic parent lookups in the future.

PR #401, opened 2020-01-23
Add warning about test timezones

loganvolkers:
Tests fail when run in the PST timezone (America/Vancouver).

I've added a note about timezones to contributing because there's probably a bigger challenge of investigating and patching the core or test suite for the underlying issue.


jhorbulyk:
I think this problem might be localized to timezones to the west of GMT.  The test suite runs fine when to the east of GMT.

andrew-coleman:
Thank you!

PR #405, opened 2020-01-30
Maintain singleton sequence of array

### LINKED ISSUE #399: Force nested arrays
I've had no luck figuring out this nuance. I'm trying to convert data to a timeseries format required by another system. It wants the data as an array of arrays like this: http://try.jsonata.org/BkKNErBxL
Input:
```
[
  {"epochSeconds":1578381600,"value": 3},
  {"epochSeconds":1578381700,"value": 5}
]
```
Expression: `$.[value,epochSeconds]`
Result:
```
[
  [3,1578381600],
  [5,1578381700]
]
```
This is correct, but when there is only one item in the input array the result gets flattened.
Input:
```
[
  {"epochSeconds":1578381600,"value": 3}
]
```
Desired Output:
```
[
  [3,1578381600]
]
```
What I'm getting:
```
[
  3,
  1578381600
]
```
Is there some way to force an outer enclosing array when there is only one data point? TIA

andrew-coleman:
The `[]` syntax in a path expression should output the result sequence as an array even if the sequence contains one value (singleton sequence).  However, this is not happening if that value is an array.  Instead it's just using that as the result array.  This PR fixes that, so that it will be enclosed in another array.

Resolves #399 

### DIFF
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -219,6 +219,9 @@ var jsonata = (function() {
         }
 
         if(expr.keepSingletonArray) {
+            if(!isSequence(resultSequence)) {
+                resultSequence = createSequence(resultSequence);
+            }
             resultSequence.keepSingleton = true;
         }
 
--- a/test/test-suite/groups/flattening/sequence-of-arrays.json
+++ b/test/test-suite/groups/flattening/sequence-of-arrays.json
@@ -1,0 +1,40 @@
+[
+    {
+        "expr": "$.[value,epochSeconds]",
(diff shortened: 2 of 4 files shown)


mattbaileyuk:
For later reference, with data:
```
{"epochSeconds":1578381600,"value": 3}
```
the expression `$.[value,epochSeconds][]` current produces
```
[3, 1578381600]
```
when it should (and this PR will enable it to) produce
```
[
      [3, 1578381600]      
]

PR #411, opened 2020-02-04
fix test suite to work in all timezones

### LINKED ISSUE #403: Timezones not working for `fromMillis`
It appears that the `fromMillis` is not respecting the timezone parameter. The docs say that it should default to UTC, but it appears to default to the system/browser timezone. Also you can not manually set the timezone.

Example JSONata expression: http://try.jsonata.org/rymaAsdZ8

```jsonata
[
 $fromMillis(1514808000000, 'Week: [w] of [xNn]'),
 $fromMillis(1514808000000, 'Week: [w] of [xNn]', "+0000"),
 $fromMillis(1514808000000, 'Week: [w] of [xNn]', "+0900")
]
```

Expected result
```json
[
  "Week: 1 of January",
  "Week: 1 of January",
  "Week: 5 of December"
]
```

Actual result
```json
[
  "Week: 5 of December",
  "Week: 5 of December",
  "Week: 5 of December"
]
```

This is causing tests the fail when they are run in other timezones, and likely causing incorrect results for anyone using the `fromMillis` function.

Also related to #347 and the lack of daylight-savings-time support.

**aside** I think that this should probably get patched, but also might want to direct people to using their system-native functions for formatting, like the `Intl` browser API and [Luxon](https://moment.github.io/luxon/docs/manual/why.html) or moment.js. Datetime processing is notoriously difficult to get right.

andrew-coleman:
The test suite failures are caused by careless use of non-UTC Date functions in the date/time parser

resolves #403 

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -644,7 +644,7 @@ const dateTime = (function () {
         // the week starts on a Monday - calculate the millis for the start of the first week
         // millis for given 1st Jan of that year (at 00:00 UTC)
         const jan1 = Date.UTC(ym.year, ym.month);
-        var dayOfJan1 = (new Date(jan1)).getDay();
+        var dayOfJan1 = (new Date(jan1)).getUTCDay();
         if (dayOfJan1 === 0) {
             dayOfJan1 = 7;
         }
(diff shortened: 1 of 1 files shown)


mattbaileyuk:
Note: This isn't just a testsuite update, but could change the output of time functions in certain circumstances, but will now achieved the intended outcome.

andrew-coleman:
Correct, it doesn't change the tests at all.  It fixes the code that causes the tests to fail in certain timezones.

PR #420, opened 2020-02-21
Build es5 versions of jsonata for Bower compatibility

### LINKED ISSUE #417: Bower doesn't find jsonata.js
Hello there!
In my project (Grunt+Bower ) "grunt wiredep" doesn't put the script in the index.html.
I see that in the bower.json of jsonata there is the  'main": "jsonata.js"' but there is no file jsonata.js
To use jsonata I downgraded to the 1.5.4 version.

If you want I'll be happy to compile the file and make a PR.
Just let me know :-)

fantonangeli:
Fixes #417 

Build for bower compatibility of:
jsonata-es5.js
jsonata-es5.min.js
jsonata.js


mattbaileyuk:
Thanks for the PR, but as these are generated files, created by the prepublish step, I don't think we want these checked into the repo.

I actually hadn't spotted this file which has been living unloved for 3 years. I haven't used bower in a long time; can it not pick up what's in the npm package, or run the generation; I'm surprised the output ES5 and CommonJS files would need to be in the repo.

fantonangeli:
Thanks for your answer @mattbaileyuk . Yes the prj I'm working for is quite outdated.
Is there something I can do to help with this issue?

andrew-coleman:
Hi @fantonangeli, as Matt says, we don't want to add the generated files back into github.  However, it looks like you can configure bower to use [custom resolvers](https://bower.io/docs/pluggable-resolvers/) and there is [one available](https://www.npmjs.com/package/bower-npm-resolver) to pull the package from npm.  Hope that helps with this.  Thanks!

fantonangeli:
Hi @andrew-coleman, thanks for your information. 
I will  try with with the resolvers.

PR #429, opened 2020-03-23
unpack outer wrapper in $string()

andrew-coleman:
Fixes a regression when the $string() function is applied to a top-level array causing the output to be enclosed in an additional array


mattbaileyuk:
Travis passed but for some reason didn't report back the success to GitHub, so I merged anyway.

PR #435, opened 2020-04-22
rewrite array flattening to avoid RangeError

### LINKED ISSUE #434: Path expression on very large object causes RangeError
When mapping over objects that contain a large number of properties (>> 100000), the evaluator throws the js error `RangeError: Maximum call stack size exceeded`.

This is a bug in the flattening logic caused by the use of `Array.prototype.push.apply` to join arrays and is related to the known issue https://github.com/nodejs/node/issues/27732.

This bit of code needs rewriting.

andrew-coleman:
resolves #434 

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -1671,7 +1671,7 @@ const functions = (() => {
                 var res =  lookup(input[ii], key);
                 if (typeof res !== 'undefined') {
                     if (Array.isArray(res)) {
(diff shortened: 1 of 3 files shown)


mattbaileyuk:
For future context, this is avoiding https://github.com/nodejs/node/issues/27732 whereby the use of the node.js spread function overloads the `push()` function.

PR #454, opened 2020-06-05
Adding missing type properties

myspivey:
arguments, name, procedure were missing from types


mattbaileyuk:
@myspivey Thanks for the PR!

myspivey:
No prob! I was playing some more this AM and noticed I got it a bit wrong. That union is off. Will throw another PR up shortly with how I see it working now for review.

PR #478, opened 2020-10-04
[Feature] Continuation callbacks to avoid malicious scripts 

LRagji:
Hi would like to contribute a feature which can be used to avoid infinite recursions or scripts which can potentially be used to lock up cpu..

Approach: Since the execution was already an iterator loop a callback can be invoked to check if the user wants to proceed with the next step or cancel the execution in async evaluations.

Examples: Script like `($x:= λ($c,$n,$b){ $c=$b?$n%2:$x($c+1,$floor($n/2),$b)};$x(0,number,bitIndex))` which uses recursion to find the bit at a given position can loop infinity if wrong bit index is  given  like -1, since many application using  jsonata use  it to run user provided scripts this  kind of callback will help them to track cpu time and memory as well if required thus making it a  good sandbox for such things. 

Signed-off-by: Laukik R


LRagji:
@andrew-coleman would help if this could be reviewed and merged. also i cant add reviewers

andrew-coleman:
Hi @LRagji, thank you for offering this PR, it's certainly an area that needs addressing for reasons you've identified.
I've started to build a JSONata visual debugger, which also has the requirements of breaking into expression evaluation, stepping and potentially aborting, but with the added requirement of examining values and expression position at each step.
The approach I have taken can be seen here:
https://github.com/jsonata-js/jsonata/blob/debugger/src/jsonata.js#L2146

Although your approach fits your needs perfectly, I'd rather hold off merging until I've made more progress with the debugger so we can get a consistent API for all needs.  I'll keep this PR open for reference.
Thanks!

LRagji:
@andrew-coleman Thanks for the update, it seems direction was right our paths differ,  while you are working on this i agree we should hold the PR, in the interim do we have any solution for sandboxing scripts coming from untrusted sources?, while the on going work is more in terms of Debugger, my challenge was from security perspective(and i believe others may also have the same) do you think extending the callback with context object can serve both our purpose to get this in faster?

PR #491, opened 2020-11-16
feat(regex): Add support for pluggable regex engines for use in evaluation

maxwellgerber:
*Motivation:*
I would like to use JSONata with [node-re2](https://github.com/uhop/node-re2) - a regex node.js implementation that guards against backtracking. 
This - in combination with work like https://github.com/jsonata-js/jsonata/pull/478 - will make it easier for me to run user-supplied JSONata in a sandbox. 

The specific attack I'm looking to guard against is an [ReDoS](https://nodejs.org/en/docs/guides/dont-block-the-event-loop/#blocking-the-event-loop-redos)

### DIFF
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -1088,7 +1088,7 @@ var jsonata = (function() {
      * @returns {Function} Higher order function representing prepared regex
      */
     function evaluateRegex(expr) {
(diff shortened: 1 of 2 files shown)


andrew-coleman (approved):

andrew-coleman:
👍 thanks

Issue #505, opened 2021-03-13
And  


E1cid:
hope i explain this correctly. using node-red and jsonata.

Ok expression ($type(test1) = "number" and test1 > 10 )
 test 1 = "test"
now if test is not a number, why does the expression contine to to the and condition?

This cause the jsonata to throw an error as test1 is then compared to a number.

If the first condition of an and expression is false then the others need not be tested.

If there is another way to avoid the error could you please point it out.

jhorbulyk:
Possible duplicate of #338

E1cid:
Cheers, for reply i did that work around after i posted.
Is there any plan to have the and condition return after first condition fails?

andrew-coleman:
resolved by #506

PR #506, opened 2021-03-14
Make boolean operators short-circuit

### LINKED ISSUE #338: Proposal:  Add short-circuit evaluation for conditionals
Consider the following input:
`{"foo":""}`
and the following JSONata expression:
`foo != "" and $number(foo) > 0`
If short-circuit conditionals existed in JSONata, then the result would be `false`.  However, because these do not exist in JSONata, the following error is produced:
`Unable to cast value to a number: ""`

However, currently a work around is possible with `foo != "" ? $number(foo) > 0 : false`

markmelville:
My first ever PR to an open-source project. Woo!

Fixes #338 and #505.

Defers execution of the right-hand side of boolean expressions. Uses the natural short-circuiting behavior of Javascript's `||` and `&&` operators to evaluate only when needed.

### DIFF
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -457,9 +457,21 @@ var jsonata = (function() {
     function * evaluateBinary(expr, input, environment) {
         var result;
         var lhs = yield * evaluate(expr.lhs, input, environment);
-        var rhs = yield * evaluate(expr.rhs, input, environment);
         var op = expr.value;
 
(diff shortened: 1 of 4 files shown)


andrew-coleman (approved):

andrew-coleman:
Thanks Mark!

mattbaileyuk:
@markmelville Just to echo that, thanks for the contribution, and glad for JSONata to be your first!

PR #508, opened 2021-03-25
Convert generators to async/await

avaidyam:
This should allow non-blocking/asynchronous functions (i.e. `fetch()`, `fs.readFile()`, other I/O-bound custom functions) to execute in parallel. It does significantly change the semantics of how the JSONata engine handles functions, and as a byproduct, it is no longer possible to `yield` on built-in async functions from within a user-defined generator function. Aside from that **breaking change,** I've ensured that all other test cases pass correctly. 

---

Side-note: the motivation for this stems from something similar to this fun simple example:
```js
import AWS from 'aws-sdk';
(async () => {
  let expression = jsonata(`
    $ec2.describeInstances().promise().Reservations.Instances.(
      $L := InstanceId; 
    {
      "Id": $L,
      "Type": InstanceType,
      "Volumes": $ec2.describeVolumes().promise()
        .Volumes[Attachments.InstanceId = $L].{
          "Id": VolumeId,
          "Type": VolumeType,
          "Size": Size,
          "State": State
        }
    })
  `)
  expression.assign("ec2", new AWS.EC2({ region: 'us-east-2' }))
  return await expression.evaluate({}, {})
})();
```

Which produces (in ~600ms instead of ~1100ms): 
```json
[
  {
    "Id": "i-1",
    "Type": "c5.2xlarge",
    "Volumes": [
      {
        "Id": "vol-A",
        "Type": "gp2",
        "Size": 30,
        "State": "in-use"
      },
      {
        "Id": "vol-B",
        "Type": "gp2",
        "Size": 8192,
        "State": "in-use"
      }
    ]
  },
  {
    "Id": "i-2",
    "Type": "c5d.4xlarge",
    "Volumes": {
      "Id": "vol-C",
      "Type": "gp2",
      "Size": 30,
      "State": "in-use"
    }
  },
  {
    "Id": "i-3",
    "Type": "r5n.xlarge",
    "Volumes": [
      {
        "Id": "vol-D",
        "Type": "gp3",
        "Size": 4096,
        "State": "in-use"
      },
      {
        "Id": "vol-E",
        "Type": "gp3",
        "Size": 16,
        "State": "in-use"
      }
    ]
  }
]
```

Of course, there's more mapping and filtering and parallel API calls from there.


andrew-coleman:
Many thanks for preparing this PR.  Because of the breaking change to the API (specifically the signature of the evaluate method), I'm going to hold off merging it into master until I've done some performance testing.  It may be that I will create a 2.0 branch for some future major release items including this.

PR #524, opened 2021-06-03
Add missing properties to ExprNode type definition

saswatds:
While working with the library I encountered the following missing properties from the ExprNode type definition.
* `expressions` exists when type is `block`
* `lhs` and `rhs` exist when type is `binary`


andrew-coleman (approved):

andrew-coleman:
Thank you!

PR #539, opened 2021-08-31
[Feature] Added minimal cli

jtara1:


jtara1:
we can test this by checking out my branch, installing the module locally like `npm i -g .`, `jsonata my-file.json SomeExpression` or just have the only positional arg be the <expression> and pipe or redirect the file contents

PR #545, opened 2021-10-15
toMillis: Allow width formatting without separators

JAMSUPREME:
# Bug Summary

Width formatting doesn't appear to be respected without a separator. Corresponding issue: https://github.com/jsonata-js/jsonata/issues/546

# Versions

JSONata `1.8.5`
Node `v14.17.0`

# Detailed Information

Reference spec for width formatting: https://www.w3.org/TR/xpath-functions-31/#date-time-examples

At the moment, it appears the width formatting is not honored within `$toMillis` due to the way the regex is being generated. For an example input `201802`, I should be able to specify a matcher `[Y0001][M01]` which will result in a parsed value of `2018/02/01`. At the moment, the library incorrectly grabs too many characters due to the `+` regex so I end up with `20180/02/01`

## Test Cases

| Input | Picture | Output (timestamp) | Output (pretty) |
| --- | --- | --- | --- |
| 201802 | `[Y0001][M01]` | 1517443200000 | Thu, 01 Feb 2018 00:00:00 GMT |
| 201802 | `[Y,*-4][M01]` | 1517443200000 | Thu, 01 Feb 2018 00:00:00 GMT |
| 20180205 | `[Y0001][M01][D01]]` | 1517788800000 | Mon, 05 Feb 2018 00:00:00 GMT |

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -983,6 +983,7 @@ const dateTime = (function () {
                         return offsetHours * 60 + offsetMinutes;
                     };
                 } else if (part.integerFormat) {
+                    part.integerFormat.n = part.n;
                     res = generateRegex(part.integerFormat);
                 } else {
                     // must be a month or day name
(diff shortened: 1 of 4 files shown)


andrew-coleman (approved):
Looks good.  Many thanks!

Issue #547, opened 2021-10-30
An error occurred when convert empty array to an object in 1.8.5


hankei6km:
I'm trying to convert an array of API responce to an object with JSONata.
in many cases it works very well.
(JSONata is very powerful)

However, I get an error when I convert an empty array in JSONata 1.8.5.
(in 1.7.0 work as expected)

```
Key in object structure must evaluate to a string; got: undefined
```

## steps to reproduce

I tried it with JSONata Exerciser.

### array to object

It works in the following cases.

https://try.jsonata.org/uvEJnSWm-

jsonata: `$${id:{"label":label,"value":value}}`

json:
```json
[
  {
    "id": "abc",
    "label": "ABC",
    "value": 123
  },
  {
    "id": "def",
    "label": "DEF",
    "value": 321
  }
]
```

result:
```
{
  "abc": {
    "label": "ABC",
    "value": 123
  },
  "def": {
    "label": "DEF",
    "value": 321
  }
}
```

### empty array

An error occurred when json was an empty array.
https://try.jsonata.org/r9UY9cPmx
(in 1.7.0 work as expected)

json:
```json
[]
```

result (1.8.5):
```
Key in object structure must evaluate to a string; got: undefined
```

result (1.7.0):
```
{}
```

## workaround

The following query works even with an empty array

```
$count($$) > 0 ? $${id:{"label":label,"value":value}} : {}
```

hankei6km:
Additional Information.

In "1.8.0 - 1.8.4" worked as expected.


## steps to reproduce

```
$ node --version
v14.18.1
```

```javascript
import jsonata from 'jsonata';

const data = [];
try {
  const expression = jsonata('$${id:{"label":label,"value":value}}');
  console.log(
    `result => ${JSON.stringify(expression.evaluate(data), null, '  ')}`
  );
} catch (err) {
  console.log(`err => ${err.message}`);
}

```

1.8.0 - 1.8.4
```
result => {}
```

1.8.5
```
err => Key in object structure must evaluate to a string; got: undefined
```

andrew-coleman:
Resolved by #555

asampal:
Doesn't this warrant a new release?

mattbaileyuk:
@asampal 1.8.6 was released today with this fix (among others)

PR #555, opened 2021-12-10
Fixing reported issue https://github.com/jsonata-js/jsonata/issues/547

kexuibm:
@andrew-coleman  this PR fixes an issue reported in https://github.com/jsonata-js/jsonata/issues/547 where jsonata 1.8.5 throws an error(**Key in object structure must evaluate to a string; got: undefined**) when an empty array is passed in.

Signed-off-by: Ke Xu <ke.xu@ibm.com>


kexuibm:
Per request from @andrew-coleman, I added unit test coverage for the defect fix.

test/implementation-tests.js
```diff
@@ -847,6 +847,17 @@ describe("Tests that include infinite recursion", () => {
...
+    // When converting an empty array of API response to an object with JSONata
+    // the library should return an empty object instead of throwing an error
+    // https://github.com/jsonata-js/jsonata/issues/547.
+    it('should return empty object', () => {
+        var expr = jsonata("$${id:{'label':label,'value':value}}");
+        var result = expr.evaluate([]);
+        expect(result).to.deep.equal({});
+    });
+});
+
```
andrew-coleman, line 860: Many thanks for adding the test case.  But since it's not an implementation specific test, please could you add it as an implementation agnostic test case in JSON format.  I suggest adding a new file `jsonata/test/test-suite/groups/object-constructor/case026.json` containing the following:
```
{
    "expr": "$${id:{'label':label,'value':value}}",
    "data": [],
    "bindings": {},
    "result": {}
}
```

andrew-coleman (changes requested):

test/implementation-tests.js
kexuibm: Thanks @andrew-coleman I updated the PR to add this new file and removed the test case from implementation-tests.js.

src/jsonata.js
```diff
@@ -964,7 +965,7 @@ var jsonata = (function() {
                 env = createFrameFromTuple(environment, tuple);
             }
             var value = yield * evaluate(expr.lhs[entry.exprIndex][1], context, env);
-            if(typeof value !== 'undefined') {
+            if(typeof value !== 'undefined' && key !== 'undefined') {
```
andrew-coleman, line 968: Ah, I've just noticed that you're comparing with the string 'undefined' rather than the undefined type.  This will break any JSON object that has a property name 'undefined' (it could happen!).
E.g. the following will break under this PR: https://try.jsonata.org/3KdIrw3DC

andrew-coleman (changes requested):

src/jsonata.js
kexuibm: Thanks @andrew-coleman that is a good point.  I refactored the code to just check the key for the undefined type and removed the string check.

andrew-coleman (approved):

PR #566, opened 2022-02-26
Update numeric-functions.md

HaroldPetersInskipp:
Correct the result of the first example for $power()


mattbaileyuk:
Thanks for spotting the error and the PR. At the moment this will just change the docs for 1.7.x, could you also please update [the unversioned docs file](https://github.com/jsonata-js/jsonata/blob/master/docs/numeric-functions.md) so the change goes into future versions.

I can then follow up with a 1.8.x versioned file (it needs to be copied from the unversioned one to [here](https://github.com/jsonata-js/jsonata/tree/master/website/versioned_docs/version-1.8.0)) which I've been meaning to do for another fix.

PR #570, opened 2022-04-12
Fix error position property for errors in functions

asennikov:
While playing with the JSONata exerciser, I've noticed that the `position` property gets set correctly for all of the errors I was able to reproduce, except for everything related to calling functions, see a couple of examples below.

looks good:

image

looks not good:

image
image

It seems like due to how JSONata evaluates functions, the position is already moved by 1 to the opening bracket character, which messes up the error highlighting, as the `error.token` does not include the opening `(` character, but the position is already past it.

IMHO, it would be better if JSONata was reducing the position by 1 to match the reported token. 

This PR is aiming to implement this adjusted behavior, but if there's a better way to solve this problem - I'm open to feedback!

### DIFF
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -1412,7 +1412,7 @@ var jsonata = (function() {
             throw {
                 code: "T1005",
                 stack: (new Error()).stack,
-                position: expr.position,
+                position: expr.position - 1,
                 token: expr.procedure.steps[0].value
             };
         }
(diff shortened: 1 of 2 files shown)


mattbaileyuk:
I am concerned that this would be a breaking change, altering the behaviour for anyone who has been an interface that relies on the current error output; the exerciser is just one example here, and I suspect this might impact IBM App Connect (where JSONata originated), and I'm aware there are a number of UIs now built which use the language.

I'm not saying we _shouldn't_ make this change, just that it might need some caution/major version bump.

asennikov:
> the exerciser is just one example here

yeap, that's fair - I mostly used exerciser as an easy example to reproduce the issue, but I'm actually more interested in fixing it for the product I'm working on which exposes JSONata to our users and displays these errors.

PR #578, opened 2022-06-07
chore: added performance benchmark test

tlakomy:
Screenshot 2022-06-07 at 17 03 26

This currently takes ~2000ms to pass, the plan is to optimize JSONata so that this particular test (and by extension - all JSONata expresions) are significantly faster.


tlakomy:
This PR was meant to be created in https://github.com/Stedi/jsonata, apologies for the noise.

Issue #568, opened 2022-03-07
`$map` expression behaves differently depending on whether it's used inside of an object or not


tlakomy:
Hello - first of all, thank you for making JSONata, it's absolutely excellent.

I've noticed an issue which seems to be a bug (reproducible in both JSONata 1.7.0 and 1.8.6, tested in https://try.jsonata.org/ explorer).

**Steps to reproduce:**
- Use a following input sample JSON file (this issue occurs for any input array):
```json
[
  1,
  2,
  3,
  4,
  5
]
```
- Write a JSONata expression that maps through a hardcoded array of two items and stores the result in `"item"` field.
```json
{
    "item": $map([1, 2], function($number, $index) { { "number": $number, "index": $index } }).{"foo": "bar"}
}
```

Expected result:

```json
{
  "item": [
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
  ]
}
```

`$map`ping over two items should produce two items at the output.

Actual result:
```
{
  "item": [
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    },
    {
      "foo": "bar"
    }
  ]
}
```

**Notes:**
- This issue occurs if instead of a hardcoded array (e.g. `[1,2,3]`) an array obtained as a result of JSONata expression is used (e.g. `Order.Products`)
- This issue **does not occur** if a `$map` expression is **not** assigned to a field inside of an object, for a following JSONata expression:

```
$map([1, 2], function($number, $index) { { "number": $number, "index": $index } }).{"foo": "bar"}
```

The result is:

```json
[
  {
    "foo": "bar"
  },
  {
    "foo": "bar"
  }
]
```

which is to be expected.

cthorner:
I’m running into a basic array navigation issue:
```
Input: [{"a": 1},{"a": 2},{"a": 3}]
Expression: $.a[1]
Result: ** no match **

Expected Result: 2
```
Here is the Exerciser link: https://try.jsonata.org/Rs_HjiRXm
It looks like the result array lives in 
```
$.a[0] and $.a[0][0][0][0][0]
```

accessing a directly solves the issue, as in:

```
a[1]
```

however, if we change the input slightly:

```
[...]

PR #579, opened 2022-06-08
fix: fixed using map operator over the result of an array higher-order function

tlakomy:
**Steps to reproduce:**
Take this expression as an example (link: https://try.jsonata.org/l340GmO68)
```
{ 
  "item": $map([1, 2], function($number) { $number }).{"foo": true}
}
```

`$map` is supposed to map over the array of two items, but given an input containing N items, e.g.
```ts
[
  1,
  2,
  3,
  4,
  5,
  6,
  7
]
```
The output contains N * M items (where M is the number of items in the first argument of `$map` function). Note that the JSONata expression does not depend on anything in the input and as such input should be completely irrelevant here.

This issue **does not** occur when using `$map`/`$filter` (or any other higher-order array function) outside of an object, e.g.

```
$map([1, 2], function($number) { $number }).{"foo": true}
```
produces two items at the output regardless of an input.

---
**Resolution:**
This PR introduces a fix for this issue (see tests for details, happy to add more test cases if needed). Note that this behaviour is a breaking change since it's possible (although IMO unlikely) that there were JSONata use cases depending on this behaviour.

All other tests are successfully passing:

```bash
  3324 passing (10s)

[...]
=============================== Coverage summary ===============================
Statements   : 100% ( 3468/3468 ), 2 ignored
Branches     : 100% ( 1990/1990 ), 10 ignored
Functions    : 100% ( 295/295 )
Lines        : 100% ( 3450/3450 )
================================================================================
```
---

Closes https://github.com/jsonata-js/jsonata/issues/568


andrew-coleman:
Thanks for raising this.  I agree that there is an inconsistency here and I agree that this would be classified as a bug fix rather than a breaking change.  But I don't understand the logic behind the code change and I wonder if you can explain your thinking given the following observation:

You are forcing the input array to be treated as a tuple when, in fact, it is not.  Tuples are an internal mechanism that is used to process expressions that contain the positional and context binding operators (`#` & `@`), and your code change takes it down the path where it tries to create an environment frame that contains these bindings, resulting in an undefined context (because it's not a tuple).  It feels to me that this fixes your original test case more by accident, which is not going to be good for maintainability when the next issue in this area arises.  

Thanks! :)

tlakomy:
Hello @andrew-coleman - thank you for your answer and insight on how JSONata tuples work under the hood.

I'll do my best to come up with a better solution (without an undefined context), would you be able to give me some pointers? I'd definitely would rather not submit fixes that work by accident/are difficult to maintain in the long run.

PR #582, opened 2022-06-21
Enable faster JSONata evaluation by switching from generators to async/await

tlakomy:
Description: TBD


tlakomy:
This PR was meant to be created inside of a fork, apologies for the noise.

PR #623, opened 2023-03-13
Fix regex termination lexer

Ben-Steele:


andrew-coleman:
Thank you for the contribution!

Issue #370, opened 2019-09-30
Proposal: Add a 'default' operator


jhorbulyk:
In JavaScript, it is possible to write statements such as:
`value = input || 0;`
where a variable or result can be assigned the result of another variable if it is assigned and otherwise it is assigned a default value.  The closest thing in JSONata that can do this is:
`$value := input ? input : 0;` 
which is significantly more verbose.  Ideally, it would be nice if there was a way to do this in JSONata.  One possibilty is to add a new operator (e.g. `||`) or is to tweak the existing `or` operator.

markmelville:
This would be a wonderful addition to Jsonata. `||` is the correct JavaScript idiom, but other options could be considered, such as `?:` or `??`, as borrowed from other languages.

To expound on the OPs example of how verbose workaround can get, if you have a large expression, you don't want to have to repeat it in a ternary:
```
$lookup($lookup($collections,$collectionName),$key) ? $lookup($lookup($collections,$collectionName),$key) : 'fallback'
```
the most terse way to do it now is with statements:
```
($value:=$lookup($lookup($collections,$collectionName),$key); $value ? $value : 'fallback')
```
with a new operator it could be:
```
$lookup($lookup($collections,$collectionName),$key) || 'fallback'
```
Another workaround is a method, but an operator would be better to short-circuit the potentially expensive work of resolving the second operand:
```
$coalecse($value,$expensiveFallback())
```

andrew-coleman:
Just want to clarify desired behaviour here.  The expression:
`$input ? $input : "fallback"`
will evaluate the condition by casting the value of `$input` to a Boolean as if by applying the [$boolean()](https://docs.jsonata.org/boolean-functions#boolean) function.  If it evaluates to `true`, then the conditional expression will return the value of `$input`, otherwise it will return the string `"fallback"`.  So, if `$input` contains the empty sequence, empty array, empty object, empty string, zero or `null`, then the expression will return `"fallback"`.  Is that the behaviour you are looking for with this 'default' operator?

Alternatively, you might want to default to a fallback value only if the `$input` is non-existent (empty sequence).  In this case you can take advantage of [sequence flattening](https://docs.jsonata.org/processing#sequences) to do the 'coalesce'.  For example:
`[$input, $input2, 'fallback'][0]`
[...]

PR #638, opened 2023-06-08
added default (elvis) operator

markmelville:
Implements request from #370.

I have never been sure about how the project owners feel about the addition of operators. Functions are one thing, but an operator is a first-class citizen, and eating one up has roadmap-altering consequences. Is there even a roadmap?

Of all the proposed syntaxes, I opted to use the elvis operator `?:` since it's so close to its functional equivalent.
`property ? property : 'default'` becomes `property ?: 'default'`. (Just collapse the `?` and the `:` of a ternary.)

Draft mode for now since:
- I want to get general input about the direction and use of a new operator
- Right now the implementation is pure javascript by using the javascript ternary, which may have different rules for what is "truthy" than the JSONata ternary. I suspect this should change to match JSONata's ternary logic.
- I'm wondering if this operator should be a short-circuiting operator to prevent the overhead of the evaluation of a potentially expensive RHS that won't be used. Thoughts?
- Documentation needs to be updated.

Thanks for looking!

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -40,6 +40,7 @@ const parser = (() => {
         '<=': 40,
         '>=': 40,
         '~>': 40,
+        '?:': 40,
         'and': 30,
         'or': 25,
         'in': 40,
@@ -197,6 +198,11 @@ const parser = (() => {
                 // ~>  chain function
                 position += 2;
                 return create('operator', '~>');
+            }
+            if (currentChar === '?' && path.charAt(position + 1) === ':') {
+                // ?: default / elvis operator
+                position += 2;
+                return create('operator', '?:');
             }
             // test for single char operators
             if (Object.prototype.hasOwnProperty.call(operators, currentChar)) {
@@ -565,6 +571,7 @@ const parser = (() => {
         terminal("in"); //
         prefix("-"); // unary numeric negation
         infix("~>"); // function application
+        infix("?:"); // default value
 
         infixr("(error)", 10, function (left) {
             this.lhs = left;
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -482,6 +482,9 @@ var jsonata = (function() {
                 case '>=':
                     result = evaluateComparisonExpression(lhs, rhs, op);
                     break;
+                case '?:':
+                    result = evaluateDefault(lhs, rhs, op);
+                    break;                    
                 case '&':
                     result = evaluateStringConcat(lhs, rhs);
                     break;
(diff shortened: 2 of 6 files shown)


markmelville:
Ah. I meant to open this in draft mode, and I can't seem to change it now.

src/parser.js
```diff
@@ -40,6 +40,7 @@ const parser = (() => {
         '<=': 40,
         '>=': 40,
         '~>': 40,
+        '?:': 40,
```
markmelville, line 43: any input on what the binding power should be on this operator?

andrew-coleman:
Hi Mark, many thanks for taking the lead on this, I’m happy that we get this into the language.  Re-reading the discussion, I think there are two separate operations that we could add:

- Elvis operator, as you propose - `A ?: B` which is equivalent to the existing `A ? A :B`.
- Coalescing operator `A ?? B` which will evaluate to B iff A does not exist (empty sequence).  Equivalent to `[A, B](0)` (from XPath).

As far as implementation is concerned, it would probably be easiest to modify the parser to check whether the next token is `:` (or `?`) before attempting to parse the ‘then’ expression https://github.com/jsonata-js/jsonata/blob/master/src/parser.js#L843
Then amend the evaluator to handle the different usages https://github.com/jsonata-js/jsonata/blob/master/src/jsonata.js#L1067

On the question of roadmap, I don’t have any committed plans at the moment, but I’m having thoughts in the following areas:

- A simpler syntax for accessing property names in objects.  I’m seeing lots of questions in Slack and StackOverflow (thank you Mark for answering many of them) along the lines of “How to I transform `{“A”: “B”, …}` into `{“key”: A, “value”: B}`".  At the moment, the only answer is to use the $each function with a lambda function to generate the output.  Which is a bit long winded for simple use cases.  XPath has a `name()` function (https://www.w3.org/TR/xpath-functions-31/#func-name) which I didn’t implement in JSONata because unlike XML, values are just values; they’re not owned by a DOM node with a name.  But I think I need to revisit how to implement something like this.
- I’ve never liked the syntax for defining lambda functions.  I don’t ever plan to make breaking changes, but it’s interesting to see that a proposal for XPath 4.0 has added support for arrow functions https://www.saxonica.com/qt4specs/XP/xpath-40-diff.html#id-inline-func.  I might look at doing something similar. 

Neither of these would affect your proposal though.

jhorbulyk:
> I think there are two separate operations that we could add:
>
>    * Elvis operator, as you propose - A ?: B which is equivalent to the existing A ? A :B.
>    * Coalescing operator A ?? B which will evaluate to B iff A does not exist (empty sequence). Equivalent to [A, B](0) (from XPath).


I agree. The A?:B is nice in that it simplifies many expressions written as A ? A : B but it might slightly miss the actual use case of "we want to provide a default value when there is none". However, we can start with this operator and see where we end up.

PR #639, opened 2023-06-12
Update overview.md

jt-nti:
The developerWorks video does not exist anymore :(


mattbaileyuk (approved):
Thanks for the contribution, even if it's an unfortunate one 😞

Issue #179, opened 2018-03-29
Test suite relies on JavaScript object traversal


deepilla:
I'm running the Go port of JSONata against the jsonata-js test suite and a few tests are failing because of assumptions that hold true in JavaScript but not in Go. As the test suite aims to be language-agnostic, it would be good to resolve these issues.

#### The Problem

The underlying problem is object traversal. When you iterate over an object in JavaScript, the values come back in source order. This code, for example, always outputs "a", "b", and "c" in that order.

``` JavaScript
var obj = {"a": 1, "b": 2, "c": 3};

for (var key in obj) {
    console.log(key);
}
```

In Go, object traversal is randomised. The equivalent Go code returns the same values, but the order is scrambled each time it runs.

``` Go
obj := map[string]int{"a": 1, "b": 2, "c": 3}

for key := range obj {
    fmt.Println(key)
}
```

This isn't an issue in JSONata itself. The [JSON spec](http://json.org/) defines an object as an **unordered** list of name/value pairs and that's exactly how Go maps work. However some of the tests in the test suite rely on JavaScript-style object traversal.

#### Affected Tests

The following tests all traverse objects and put the items into an array. The resulting array is assumed to be in a particular order. This order is not guaranteed in Go (or implied by the JSON spec).

- [`$each(Address, λ($v, $k) {$k & ": " & $v})`](https://github.com/jsonata-js/jsonata/blob/master/test/test-suite/groups/function-each/case000.json)
- [`$keys(Account)`](https://github.com/jsonata-js/jsonata/blob/master/test/test-suite/groups/function-keys/case000.json)
- [`$keys(Account.Order.Product)`](https://github.com/jsonata-js/jsonata/blob/master/test/test-suite/groups/function-keys/case001.json)
- [`$spread((Account.Order.Product.Description))`](https://github.com/jsonata-js/jsonata/blob/master/test/test-suite/groups/function-spread/case001.json)
- [`foo.*`](https://github.com/jsonata-js/jsonata/blob/master/test/test-suite/groups/wildcards/case000.json)
- [`*[type="home"]`](https://github.com/jsonata-js/jsonata/blob/master/test/test-suite/groups/wildcards/case007.json)

The following tests do the same thing but add a predicate to retrieve a specific item from the array. Again, this relies on the items being in a known order.

- [`foo.*[0]`](https://github.com/jsonata-js/jsonata/blob/master/test/test-suite/groups/wildcards/case006.json)
[...]

PR #665, opened 2024-01-03
Fix binding operation evaluation in objects

asennikov:
The issue was originally raised here: https://stackoverflow.com/questions/77742737/jsonata-transform-operator-unexpected-change-in-variable-binding-between-versi

Before https://github.com/jsonata-js/jsonata/pull/583 a JSONata user could bind a variable within an RHS expression of one of the object entries and then reuse that variable within another object entry, e.g.:

```
{ 
    "foo": $bar := "x",
    "bar": $bar
}
```

Here's the result on 1.8.6:

image

But due to the `await Promise.all` added in https://github.com/jsonata-js/jsonata/pull/583, this no longer works in 2.0.3:

image

----------
 
In this PR, I'm proposing a change to remove the `Promise.all` and iterate sequentially, which might have some performance implications for certain users of JSONata, but IMO, it would still be better to maintain compatibility between versions.

All tests are passing:

image


efhjones (approved):
👍 Seems reasonable to me

scarabaeus (approved):
Thanks for the fast turn-around on this @asennikov!

VesaliaBricks:
As the originator on SO, thank you for clearing that up. Will try to re-formulate my business logic.
Bit frustrating, but I still love JSONata,
Sebastian

asennikov:
> Sorry, no.  There is no guaranteed order of execution of expressions within an object (`{...}`).  Although this Javascript implementation executes them in the same order as appears in the expression, other implementations (such as `go`) don't necessarily. See https://github.com/jsonata-js/jsonata/issues/179 for example.
> The only construct in JSONata that guarantees order of execution is the block `(... ; ... ; ...)` expression.

Sounds reasonable.

PR #685, opened 2024-03-05
fix: release take 2

nddery:
I had not authorized the token against SSO, which caused failures.


nddery:
Oops, sorry, I mistakenly opened this against the wrong repository.

PR #686, opened 2024-03-08
Proposal for a new function defaultValue()

marcioandrada:
The goal of this function is to return a default value in case the first argument is null.

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -2055,6 +2055,20 @@ const functions = (() => {
         return result;
     }
 
+    /**
+     * Returns a default value in case the first argument is null
+     * @param {object} arg1 - the object check for null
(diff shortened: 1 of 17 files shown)


jhorbulyk:
Have you seen the conversation in #370 ?

marcioandrada:
> Have you seen the conversation in #370 ?

Hello, @jhorbulyk,

No, indeed I didn't see that discussion. It is clear that other people are already addressing the same topic.  :-)   I am cancelling the Pull Request, then, since it doesn't make sense to add one more thread to something already in progress.

Thank you for the feedback.

Kind regards,
Márcio Andrada

marcioandrada:
Discarding the Pull Request, since there are other threads to address the same issue.

PR #687, opened 2024-03-14
Repair bad logic around `undefined != <anything-else>`

iiian:
`undefined != "foo"` should evaluate to `true`


iiian:
I'll try and fix tests for this later. Would love to see this bug fixed or documented. I still think tristate logic is a smell and an antipattern for this language. Am open to hearing dissenting opinions and willing to be converted.

iiian:
I dealt with my problem by just using a `Proxy` to return a `null` if my context object doesn't contain a key I'm looking for.

PR #701, opened 2024-06-08
Add undocumented/private API to hook into when a new frame is created.

adamscybot:
This is a small change to add a new hook, similar to `__evaluate_entry` and `__evaluate_exit`, but for when a new frame is created from an old one. I am building an open-source pluggable framework around JSonata, and this small change will open up an enormous amount of awesome possibilities.

It is beneficial for advanced cases like modifying env with additional metadata or mutating it so that you can analyse bind/lookup calls & provide diagnostics. I have actually _patially_ achieved this already to significant effect by hooking into the env on `__evaluate_entry`. However, in some cases inside the JSONata core, a new frame is created, which is then immediately bound to, and consequently that happens before it has yet to recurse through `evaluate` and the entry callback. That means that callback is unable to capture everything for these advance pro-mode use cases. This new callback would enable that possibility.

I accept this will be undocumented and could change in the future.

The callback is set via a `Symbol`, so it can not be manipulated or read inside a query. See #700, which would need to be merged first, both as approval of the general `Symbol` approach and because the minor TS def changes are needed from there.

Signed-off-by: Adam Thomas [adam@adam-thomas.co.uk](mailto:adam@adam-thomas.co.uk)

### DIFF
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -1837,7 +1837,7 @@ var jsonata = (function() {
      */
     function createFrame(enclosingEnvironment) {
         var bindings = {};
-        return {
+        const newFrame = {
             bind: function (name, value) {
                 bindings[name] = value;
             },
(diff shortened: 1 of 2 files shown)


shivanshu-netomi (approved):

mattbaileyuk:
All looks good, including the protection around its use, so will merge this 👍

mattbaileyuk:
Realised that didn't actually run through the checks; usually get an option on the merge to kick them off if they're initiated from a fork. Thankfully all looks good: https://github.com/jsonata-js/jsonata/actions/runs/12076711571/job/33678469127

PR #713, opened 2024-09-18
Update Java implementation version

uw4:
Impl is compatible with the latest 2.0.5 features


mattbaileyuk (approved):
Thanks. I'll have a look later this week to see about getting the latest doc changes live under the current version

PR #714, opened 2024-09-23
Chain operator should respect array constructor

netomi:
This PR fixes the problem where when using a HOF like map with a chain operator, the array constructor is not respected which is added to enforce that the output is always an array regardless of the input.

Take this example:

```
$map([1], function($v){$v*$v})[]
```

returns correctly `[1]`. However, when using a chain operator like that:

```
$ ~> $map(function($v){$v*$v})[]
```

and providing `[1]` as input, the singleton value is returned.
This PR fixes that and adds 2 test cases to the test suite.


mattbaileyuk (approved):
Thanks for fix!

PR #729, opened 2025-01-14
$pad function to truncate fractional part of numeric argument

### LINKED ISSUE #717: $pad Function With Decimal-Value Width Results in Uncaught RangeError
Hello,

Attempting to invoke `$pad` with a decimal value length like `2.1` results in `RangeError: invalid array length` being thrown.  This is due to the padding array being instantiated with a decimal value length [which is not allowed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Array#exceptions).  This exception is not caught anywhere in the call stack.

We should either floor `width` before using it to calculate `padLength` or throw a JSONata exception.

Repro: `$pad("string", 2.1, " ")`

andrew-coleman:
Truncating the fractional part of the argument is consistent with the behaviour of other functions expecting integer arguments (e.g. substring)

Resolves #717

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -293,6 +293,7 @@ const functions = (() => {
         }
 
         var result;
+        width = Math.trunc(width);
         var padLength = Math.abs(width) - length(str);
         if (padLength > 0) {
             var padding = (new Array(padLength + 1)).join(char);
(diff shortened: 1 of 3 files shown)


mattbaileyuk (approved):
LGTM

PR #731, opened 2025-01-15
Add function $uuid()

andrew-coleman:
Generates a random UUID (version 4)

Uses the crypto.randomUUID() implementation which has been available in node since v16.  Hence, removing v12 & v14 from the CI (these are long out of service anyway).


andrew-coleman:
Doesn't like node 16 either (also out of service)

mattbaileyuk:
Node.js 16 went out of support last April, and was a reason for a lot of npm modules doing major version bumps through the summer as they also removed support. We should also be doing a major version bump if we're dropping support for older versions, as per semver (it's a breaking change).

Node.js 18 is still in support until end of April this year, so would be reluctant to want to drop it at this point.

andrew-coleman:
> Node.js 18 is still in support until end of April this year, so would be reluctant to want to drop it at this point.

I agree, and as I've just discovered, this new function was included in Node from v19.  So I think we should hold off from adding this at the moment.

https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto

markmelville:
Would you consider writing your own uuid function, so that you don't need to use this crypto library? AFAIK, there's no actual cryptography to it, only random number generation and bitwise operators. When I first found jsonata, I implemented UUID 4 with a jsonata expression:
```
(
  $binPad:=function($n, $len){$pad($formatBase($n,2),-$len,'0')};
  $bitwise:=function($lb, $rb, $fn){$split($lb,'')~>$map(function($c, $i){$fn($c='1',$substring($rb,$i,1)='1')?$power(2,(7-$i)):0})~>$sum()};
  $and:=function($l,$r){$bitwise($binPad($l,8),$binPad($r,8),function($1,$2){$1 and $2})};
  $or:=function($l,$r){$bitwise($binPad($l,8),$binPad($r,8),function($1,$2){$1 or $2})};  
  [0..15]
     ~>$map(function(){$floor($random()*256)})
      ~>$map(function($v,$i){$i=6?$or($and($v,15),64):$i=8?$or($and($v,63),128):$v})
      ~>$map(function($n,$i){$pad($formatBase($n,16),-2,'0') & ($i in [3,5,7,9] ? '-' : '')})
      ~>$join()
)
```

andrew-coleman:
Thanks @markmelville, yes I started off with this approach, but the specification for UUIDv4 says that it **should** use a cryptographically secure random number generator, which led me to the `crypto` package anyway.

https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-4
https://datatracker.ietf.org/doc/html/rfc9562#unguessability

PR #745, opened 2025-05-12
Implement support of spread operator

ericleib:
This Pull Request adds support of the spread operator `...` in the Jsonata syntax.

The implementation works the same way as in Javascript:

```js
{
   "a": "hello",
   "b": "world",
   ...foo
}
```

This copies the key/value pairs from object `foo` to a new object, in this case potentially overriding properties `a` and `b`. Conversely, if `a` and `b` were added AFTER `foo`, then their values would override the `a` and `b` entries of `foo`.

Please let me know if this is something that the community would be interested in. In my case (an avid Jsonata user), this feature would make a big difference.


andrew-coleman:
Thank you for raising this.

On the face of it, this looks like a great proposal.  What's concerning me though is the implied order of execution of the items within the object.  Although not a problem for this Javascript implementation, it could be a problem for other implementations.  Take a look at [this issue](https://github.com/jsonata-js/jsonata/issues/179) for reference.  Although that issue was only relating to the test suite, I think this proposal would bring it into the language itself.  

I'm not saying that a feature like this is not possible, but we need to be very mindful of this potential problem.

ericleib:
Thank you for looking into it!

I understand your concern. It does bring a change to the language. In fact when implementing it, I could see the potential inconsistency of behavior with this error "Multiple key definitions evaluate to same key" (which prevents any key overriding mechanism).

PR #747, opened 2025-05-24
Await array elements

aslakhellesoy:
First off - thanks for this amazing library! ❤️

I've got some input with arrays of promises. Accessing properties of these array elements results in `undefined`.

This PR fixes this. I've added a test and all other tests are green.


andrew-coleman:
@mattbaileyuk are you happy with this?

PR #748, opened 2025-07-01
feat: add default/elvis and coalescing operator

zirkelc:
This PR continues the work of @markmelville in PR https://github.com/jsonata-js/jsonata/pull/638 for the elvis/default operator `A ?: B` and adds the coalescing operator `A ?? B` as described by @andrew-coleman 
> Coalescing operator `A ?? B` which will evaluate to B iff A does not exist (empty sequence). Equivalent to `[A, B](0)` (from XPath).
> https://github.com/jsonata-js/jsonata/pull/638#issuecomment-1584301119

I followed the logic of `$exists()` when evaluating the LHS which only checks for `undefined`, that means other falsy values like `null` are evaluated as true. This could be changed to also include `null` to match the behavior of JavaScript's `??` operator.

### DIFF
--- a/docs/other-operators.md
+++ b/docs/other-operators.md
@@ -27,7 +27,7 @@ __Example__
 
 ## `?:` (Default/Elvis)
 
-The default (or "elvis") operator returns the left-hand side if it is truthy, otherwise it returns the right-hand side. This is useful for providing fallback values when an expression may evaluate to a falsy value (such as `null`, `false`, `0`, `''`, or `undefined`).
+The default (or "elvis") operator returns the left-hand side if it has an effective Boolean value of `true`, otherwise it returns the right-hand side. This is useful for providing fallback values when an expression may evaluate to a value with an effective Boolean value of `false` (such as `null`, `false`, `0`, `''`, or `undefined`).
 
 __Syntax__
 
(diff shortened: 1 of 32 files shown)


markmelville:
Thanks for picking this up. When the feedback on my PR was that the `??` was probably the more important to prioritize, I admittedly lost my steam. My feedback here is that the elvis operator should get some more test cases to show that that rhs would be selected for defined, falsy values such as 0 or "" (where as the null coalescing would not).

src/parser.js
```diff
@@ -565,6 +577,8 @@ const parser = (() => {
         terminal("in"); //
         prefix("-"); // unary numeric negation
         infix("~>"); // function application
+        infix("?:"); // default value
+        infix("??"); // coalescing operator
```
andrew-coleman, line 581: Slightly more involved, but I think we can do the same here, since this is syntactic sugar for `$exists(a) ? a : b`:
```
infix("??", operators['??'], function (left) {
    this.type = 'condition';
    this.condition = {
        type: 'function',
        value: '(',
        procedure: {type: 'variable', value: 'exists'},
        arguments: [left]
    };
    this.then = left;
    this.else = expression(0);
    return this;
});
```
Then there is nothing extra to do in the evaluator.

docs/other-operators.md
```diff
@@ -25,6 +25,38 @@ __Example__
 
 `Price < 50 ? "Cheap" : "Expensive"`
 
+## `?:` (Default/Elvis)
+
+The default (or "elvis") operator returns the left-hand side if it is truthy, otherwise it returns the right-hand side. This is useful for providing fallback values when an expression may evaluate to a falsy value (such as `null`, `false`, `0`, `''`, or `undefined`).
```
andrew-coleman, line 30: This is correct, although I use the term 'effective Boolean value' rather than truthy / falsy. (this comes from the terminology used in XPath / XQuery).
Perhaps also you could add some words on this [here](https://docs.jsonata.org/next/programming#conditional-logic) too.

andrew-coleman (changes requested):
Many thanks for working on this.  I've left a few comments, but basically I think we can do this purely in the tokeniser / parser since these new operators are syntactic sugar for the existing ternary operator.

PR #751, opened 2025-07-15
feat: support trailing commas

### LINKED ISSUE #637: Support for trailing commas
**Feature request**

I would like the following code to be treated as valid by the jsonata parser, the trailing comma after `data.id` should be ignored if present.
 ```jsonata
(
  {
    "id": data.id,
  }
)
```

zirkelc:
Fixes #637 

It's probably a philosophy question whether trailing commas should be supported: they are valid in JavaScript but not valid in JSON.

I encountered many cases where I removed a line from an JSONata expression but I forgot to remove the trailing comma in the previous line and encountered an error. 
Such cases would usually be auto-fixed by the linter/formatter/IDE, but since we don't have built-in support for JSONata, I personally think trailing commas would be convenient to reduce such stylistic errors.


zirkelc:
Hey @mattbaileyuk thanks for your feedback! Yeah I expected that this might be controversial as trailing commas are not liked by everyone even in the JS world 😅 

Feel free to close the PR if you decide not to go that way! :-)

Issue #708, opened 2024-08-20
`$map` does not return an array on input arrays of length 1


vecerek:
According to the [docs](https://docs.jsonata.org/higher-order-functions#map), `$map` should return an array. However, if the input data is an array containing exactly one element, it returns the value returned by the function passed to `$map` as the second argument.

Reproduction: https://try.jsonata.org/hW13DjVzw

In case the link stops working at some point in the future, here is the inlined version of the reproduction steps:

#### Data

```json
[{ "a": 1 }]
```

#### Jsonata

```jsonata
$map($, function($el) {
  { "a": $el.a + 1 }
})
```

#### Expected output

```json
[{ "a": 2 }]
```

#### Actual output

```json
{ "a": 2 }
```

vecerek:
Temporary workaround for people struggling with this in production:

```jsonata
(
    $correctMap := function($data, $callback) {(
      $res := $map($data, $callback);

      $type($res) = "array" ? $res : [$res]
    )};

  $correctMap($, function($el) {
    { "a": $el.a + 1 }
  })
)
```

Playground link: https://try.jsonata.org/sa2yM0nj8

vecerek:
Another thing I noticed, if the input array is empty, the result is `** no match **` instead of an empty array. Are my expectations simply wrong here? The way I interpret:

> Returns an array containing the results of applying the function parameter to each value in the array parameter.

from the docs is that the function returns an array in all cases, not just sometimes.

elviejo79:
I think your intuitions are correct.
$map should return an array in all case:

empty array -> empty array
1 element array -> 1 element array
n element array -> n element array

netomi:
I also stumbled upon this weird behavior. Adding an array constructor at the end like `$map([1], ...)[]` solves it however when using the map together with a chain operator it currently does not work even with the workaround.

I created a PR to fix that also in this case to be able to use the workaround: https://github.com/jsonata-js/jsonata/pull/714

However I would like to note that the current behavior of `map` is not what you would expect and thus it can not really be used. The workaround for me is to use a lambda function that iterates over the elements of an array instead of using the built-in function map which should do the same.

If this behavior is wanted, you could add an additional built-in function to return a singleton if explicitly requested, but not do that be default imho.

vecerek:
This is also an interesting behavior. However, I'm not sure if it's directly related to this bug or it should be reported separately.

```jsonata
[
  $map($, function($el) {
    { "a": $el.a + 1 }
[...]

Issue #462, opened 2020-06-25
Inconsistent Behavior of [] between version 1.7.0 and 1.8.3


jhorbulyk:
Consider the following input:
```
{
  "singleArray": [{"foo":"bar"}]
}
```
and JSONata expression:
`singleArray[]`

In version 1.7.0, the evaluated result is 
```
[
  {
    "foo": "bar"
  }
]
```
but in 1.8.3 it is
```
[
  [
    {
      "foo": "bar"
    }
  ]
]
```
Which evaluation is correct and why?

andrew-coleman:
This is a result of this PR: https://github.com/jsonata-js/jsonata/pull/405
to fix this bug: https://github.com/jsonata-js/jsonata/issues/399

This idea is that the `[]` syntax turns the sequence into an array even if that sequence contains a single value which itself is an array.

jhorbulyk:
The problem is that I don't think that change fully works.  In JSONata 1.8.3, if we change the JSONata expression to
`(singleArray[] ~> $map(function($i){$i}))[]`
then the result is still
```
[
  {
    "foo": "bar"
  }
]
```
not
```
[
  [
    {
      "foo": "bar"
    }
  ]
]
```
and I can't find an JSONata expression that uses `$map()` to get me to the second output.

markmeeus:
The map function maps every item and returns the array of mapped items. I think the $single function will work for you here.

goulderb:
The only solution I've found that will force nested arrays when working with array functions (e.g. $map, $filter) is the following:

```
[
    [
        $map(singleArray, function($i) {$i})
    ]
]
```

Using [] at the end of the expression does not work the same as it does without functions involved. In function compositions, this means there isn't a way to force nested arrays as part of the output without wrapping with more functions to force arrays using the wrapping square brackets notation.

RedShift1:
This is stupid. Why would a function like $map() ever return anything other than an array? No other language does this.

andrew-coleman:
> This is stupid. 

Thanks for the feedback.

In common with XQuery, JSONata treats singleton sequences as equivalent to the value.

But there is potentially an issue highlighted above (slightly more eloquently) that needs investigating.

knolleary:
Just found this issue as I was expecting `$map` to always return an array. The docs for `$map` says it returns an Array and doesn't mention the singleton sequence case.

[...]

PR #753, opened 2025-07-23
Reflect missing behaviours on map in docs

AndrewLipscomb:
https://github.com/jsonata-js/jsonata/issues/462 and https://github.com/jsonata-js/jsonata/issues/708 both reflect seemingly unexpected behaviour in how map works. 

At the very least - the current behaviour has been in there for several versions now and is completely undocumented. This heavily throws new-time readers and AI tools parsing the docs. 

This is my 5 min attempt to alleviate that issue. 

I am unsure on whether `undefined` is the right term for the empty array case - potentially that needs correcting. But regardless - something is better than nothing!

### DIFF
--- a/docs/higher-order-functions.md
+++ b/docs/higher-order-functions.md
@@ -7,7 +7,31 @@ sidebar_label: Higher Order Functions
 ## `$map()`
 __Signature:__ `$map(array, function)`
 
-Returns an array containing the results of applying the `function` parameter to each value in the `array` parameter.
+If the input argument is an array of 2 or more elements, returns an array containing the results of applying the `function` parameter to each value in the `array` parameter.
+
+```
+$map([1,2,3], function($v) { $v * 2 })
+```
+
+evaluates to
+
+```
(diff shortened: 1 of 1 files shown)


docs/higher-order-functions.md
```diff
@@ -7,7 +7,31 @@ sidebar_label: Higher Order Functions
...
+$map([2], function($v) { $v * 2 })
+```
+
+evaluates to
+
+```
+4
+```
+
+If the input argument is an empty array, returns `undefined`
```
mattbaileyuk, line 34:
```suggestion
If the input argument is an empty array, returns nothing (represented in Javascript as `undefined`)
```
I agree, I think "undefined" isn't quite right here as the Javascript implementation in this repo is a reference implementation, JSONata is implemented in many languages. The existing docs refer to the empty result as "nothing" often with a note about it being `undefined` in Javascript so we could do that here too.

mattbaileyuk (approved):
Thanks for the improvement

PR #760, opened 2025-10-18
Function signature fix for repeating arguments

Turings98apprentice:
Before this change, a function signature containing a repeating symbol (eg. `<n+>`) would copy the first of those arguments into every matching parameter instead of copying each argument to each parameter respectively.

Example:
```javascript
function($a1, $a2)<n+> {
    [$a1, $a2]
}(1, 2)
```
returns `[1, 1]` instead of `[1, 2]`.

This fix modifies the behavior to match what's expected\*:
```javascript
function($n1, $n2, $n3, $s)<n+s> {
    {
        $s: [$n1, $n2, $n3]
    }
}(3, 7, 31, 'Mersenne primes')
```
now yields
```json
{
    "Mersenne primes": [
        3,
        7,
        31
    ]
}
```

_\*One important note here is that this implementation doesn't actually support an arbitrary number of arguments, only as many as the number of parameters you're willing to add to your function. I tried to change as little as possible while still squashing the bug._

---

Relatedly, I would recommend a breaking change that would bind each repeating parameter to an array of its provided arguments. Then, you could write something like this:
```javascript
function($nums, $str)<n+s> {
    {
        $str: $count($nums)
    }
}(1, 2, 3, 4, 5, 'Number of arguments')
```
which would yield
```json
{
    "Number of arguments": 5
}
```

---

Signed-off-by: Brandon Peterson <brandonman17@gmail.com>


andrew-coleman (approved):

andrew-coleman:
Thank you

Issue #761, opened 2025-11-04
`$string()` does not preserve full precision for large numeric values (≥ 15–16 digits)


MaxPlamadeala-RecoLabs:
We've been having an issue with jsonata parsing of large numbers. We had this initial json (obfuscated):
```json
"data": [
      {
        ...
        "userId": 5890840712243077,
        "createdAt": "2025-10-30T13:02:59Z",
        "modifiedAt": "2025-10-30T13:02:59Z",
      }
]
```
And this Jsonata expression:
```
$string(data[0].userId)
```

What it gives us is `"5890840712243080"` which is incorrect as you see. If I try however a smaller number like `589084071224307` (-1 digit) it works perfectly. If I try a larger nr like `589084071224307838838` it goes crazy:
<img width="265" height="100" alt="Image" src="https://github.com/user-attachments/assets/ee544cda-da19-419c-be98-d5c5e8da5ab8" />
As I see in `src/functions.js` you have [this line](https://github.com/jsonata-js/jsonata/blob/2bb258933b509b97ed173a48e0d5661033b1df19/src/functions.js#L133-L136), which cuts every number to 15 digits. Is it on purpose? Can this be fixed or do you have an analogy to propose for larger numbers?

iovanom:
I have created a fix in this PR https://github.com/jsonata-js/jsonata/pull/762

Turings98apprentice:
As I mentioned in PR #762 I think your bigger issue here is that you're using integers for user IDs instead of strings. Is there a functional requirement that the user IDs have to be integers? Do you have any control over that input JSON?

iovanom:
> As I mentioned in PR [#762](https://github.com/jsonata-js/jsonata/pull/762) I think your bigger issue here is that you're using integers for user IDs instead of strings. Is there a functional requirement that the user IDs have to be integers? Do you have any control over that input JSON?

The json we get from external API that we can't control. We can't change external API to get the string as id.

andrew-coleman:
Resolved by #762

PR #762, opened 2025-11-04
Precision fix for $string function

iovanom:
Related to https://github.com/jsonata-js/jsonata/issues/761

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -131,8 +131,10 @@ const functions = (() => {
                 arg = arg[0];
             }
             str = JSON.stringify(arg, function (key, val) {
-                return (typeof val !== 'undefined' && val !== null && val.toPrecision && isNumeric(val)) ? Number(val.toPrecision(15)) :
-                    (val && isFunction(val)) ? '' : val;
+                if (typeof val !== 'undefined' && val !== null && isNumeric(val)) {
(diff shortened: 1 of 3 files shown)


Turings98apprentice:
If you add just one more digit to the formatBase test, it fails. I think a true fix to this issue would involve incorporating something like [BigDecimal](https://www.npmjs.com/package/js-big-decimal). That being said, I think the issue this PR seeks to solve is better solved by simply using a string for large user IDs instead of integers.

image

iovanom:
Yes, it will fail, but not because of the JSONata library. JavaScript works that way. :)

You can see the example:

```js
const a = 58908407122430764;
console.log(a); // output 58908407122430770
console.log(a.toString()) // the same 58908407122430770

const b = '58908407122430764';
const c = parseInt(b);
console.log(c); // output 58908407122430770
```

The fix is not related to the `$fromBase` function; it’s related to the `$string` function. The issue with the `string` implementation isn’t due to JavaScript limitations but rather the precision parameter within the JSONata library.

I completely agree that it’s better to have a string representation for the ID, but we can’t control the third-party API that returns the ID in number format.

andrew-coleman (approved):

andrew-coleman:
Thank you!

Issue #767, opened 2026-01-25
incorrect 'each' function signature


sergeevik:
from docs:
https://docs.jsonata.org/programming
```
Invocation chaining
value ~> $funcA ~> $funcB
is equivalent to
$funcB($funcA(value))

Examples

Customer.Email ~> $substringAfter("@") ~> $substringBefore(".") ~> $uppercase()
```

in jsonata exerciser:
`$each([123, 321], function($v) {$v})` return `Argument 1 of function "each" does not match function signature`
<img width="569" height="236" alt="Image" src="https://github.com/user-attachments/assets/b29edad1-afc7-40ca-8b96-8896c5d1a83f" />

and
`[123, 321] ~> $each(?, function($v) {$v})` return `[123,321]`

<img width="453" height="321" alt="Image" src="https://github.com/user-attachments/assets/3f6ee210-82ea-4ea2-b64e-5a2e4a8cc068" />

by the docs 
`value ~> $funcA` is equivalent to `$funcA(value)` but for `each` function it's not

research this if find some strage behavior:
when each recieve array as param - it's have signature and check it. When use chaining jsonata create new runtime 'each' function without signature and not check it

in code of 'each' jsonata just iterate on key of first argument. and in js array have keys (Object.keys(arr)). and it's work 

```
    /**
     *
     * @param {*} obj - the input object to iterate over
     * @param {*} func - the function to apply to each key/value pair
     * @returns {Array} - the resultant array
     */
    async function each(obj, func) {
        var result = createSequence();

        for (var key in obj) {  <-- if obj is array code work well, but signature broken (btw not checked)
            var func_args = hofFuncArgs(func, obj[key], key, obj);
            // invoke func
            var val = await func.apply(this, func_args);
            if(typeof val !== 'undefined') {
                result.push(val);
            }
        }

        return result;
    }
```

found it's behavior when migrate my code from js to java and use https://github.com/dashjoin/jsonata-java
open issue there https://github.com/dashjoin/jsonata-java/issues/79 but it froze. 

Anyway. I suggest change 'each' signature to `<(oa)-f:a>` which allow array or object for processing. This is a more correct solution in my opinion (to avoid a kludge), and it brings the signature into line with actual functionality.

I hope you'll take a look and give me feedback, as we need to decide whether migration is possible without any kludge.

andrew-coleman:
The `$each` function is used for operating on the key/values pairs of an object: https://docs.jsonata.org/object-functions#each

[...]

PR #768, opened 2026-01-25
Issues-767: change the 'each' signature to work with arrays

sergeevik:
Change 'each' signature for correct work by docs. (https://github.com/jsonata-js/jsonata/issues/767)
Docs say: https://docs.jsonata.org/programming
```
Invocation chaining
value ~> $funcA ~> $funcB
is equivalent to
$funcB($funcA(value))
```
fix it for 'each' method
now 
`$each([123, 321], function($v) {$v})` and `[123, 321] ~> $each(?, function($v) {$v})` are equivalent 

Signed-off-by: Sergeev Viktor [sergeevviktor017@gmail.com](mailto:sergeevviktor017@gmail.com)

### DIFF
--- a/src/jsonata.js
+++ b/src/jsonata.js
@@ -1916,7 +1916,7 @@ var jsonata = (function() {
     staticFrame.bind('spread', defineFunction(fn.spread, '<x-:a<o>>'));
     staticFrame.bind('merge', defineFunction(fn.merge, '<a<o>:o>'));
     staticFrame.bind('reverse', defineFunction(fn.reverse, '<a:a>'));
-    staticFrame.bind('each', defineFunction(fn.each, '<o-f:a>'));
+    staticFrame.bind('each', defineFunction(fn.each, '<(ao)-f:a>'));
     staticFrame.bind('error', defineFunction(fn.error, '<s?:x>'));
     staticFrame.bind('assert', defineFunction(fn.assert, '<bs?:x>'));
     staticFrame.bind('type', defineFunction(fn.type, '<x:s>'));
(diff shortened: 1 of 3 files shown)


Turings98apprentice:
The [`$each` function](https://docs.jsonata.org/object-functions#each) is specifically designed for iterating over the key-value pairs of objects, so I don't think it makes sense to allow arrays to be passed in as a first argument.

This does, however, bring up an interesting point about the difference between how JSONata handles the two cases you presented, one with a literal invocation, and another with function chaining. [Example code](https://try.jsonata.org/WqQ45tZau)

I think these two tests (or something similar) should be added to the [`partial-application`](https://github.com/jsonata-js/jsonata/tree/master/test/test-suite/groups/partial-application) and/or [`function-applications`](https://github.com/jsonata-js/jsonata/tree/master/test/test-suite/groups/function-applications) test group(s) to ensure the proper type error is thrown when partially applying the `$each` function (and others).

_I'm not sure exactly what's causing these two cases to be treated differently, so more targeted tests might be a better answer here. This PR seems to point out an issue worth investigating further._

sergeevik:
@Turings98apprentice 
Although it's not intended for use with arrays, it can still be used. In my case, it's actually used, and that's just how it happened.

I don't control all the code executed by jsonata. It turns out that the differences in expected behavior (the function shouldn't be used with arrays) and actual behavior (the function is used with arrays) are causing a difference in perspective between the creators and users.

But in this case, it seems like an ideological contradiction. The author created the function for one purpose, but it performs both functions well.

I don't want to introduce new functionality, here the desire is to bring the signature to its real use, and not to its planned use

PR #774, opened 2026-03-17
Fix ?? operator with array predicate on LHS

### LINKED ISSUE #773: Nullish coalescing (??) operator returns wrong result when LHS has array predicate
## Bug Report

The `??` (nullish coalescing) operator produces incorrect results when the left-hand side contains an array subscript/predicate.

### Environment

- **jsonata version:** 2.1.0
- **Node.js version:** v24.11.0

### Steps to reproduce

Input data:
```json
{"array": [1, 2]}
```

Expression:
```
array[1] ?? 3
```

**Expected:** `2` (the second element exists, so `??` should return it)
**Actual:** `3` (incorrectly falls through to the default value)

### Root cause

The `??` infix handler in `parser.js` reuses the same `left` AST node object reference for both the `$exists()` condition check and the `then` branch. During AST post-processing, the `[` filter/predicate handler pushes filter stages onto the shared node twice, effectively turning `array[1]` into `array[1][1]` — which evaluates to `undefined`, causing the fallback to the `else` branch.

doolse:
## Summary

- Fix the `??` (nullish coalescing) operator returning incorrect results when the LHS
  contains an array predicate (e.g. `array[1] ?? 3`)
- The parser's `??` handler reused the same `left` AST node reference for both the
  `$exists()` condition and the `then` branch, causing filter stages to be duplicated
  during AST post-processing
- Deep-clone `left` for the condition argument so the two branches have independent
  AST nodes

Fixes #773

## Test plan

- [x] Added test case for LHS predicate that exists (case013)
- [x] Added test case for LHS predicate that doesn't exist (case014)
- [x] All 1763 tests passing

Signed-off-by: Jolse Maginnis <doolse@gmail.com>

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -585,7 +585,7 @@ const parser = (() => {
                 type: 'function',
                 value: '(',
                 procedure: { type: 'variable', value: 'exists' },
-                arguments: [left]
+                arguments: [JSON.parse(JSON.stringify(left))]
             };
             this.then = left;
             this.else = expression(0);
--- a/test/test-suite/groups/coalescing-operator/case013.json
+++ b/test/test-suite/groups/coalescing-operator/case013.json
@@ -1,0 +1,7 @@
+{
+    "description": "array predicate on lhs exists, should return lhs value",
+    "expr": "foo.blah[0] ?? 42",
+    "dataset": "dataset0",
+    "bindings": {},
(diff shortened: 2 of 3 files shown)


andrew-coleman (approved):

andrew-coleman:
Many thanks :)

Issue #776, opened 2026-03-29
incorrect optional arguments signature validation


mikhail-barg:
I have been updating my [C# jsonata port](https://github.com/mikhail-barg/jsonata.net.native) to pass [latest](https://github.com/jsonata-js/jsonata/commit/d4e4133445357286288c5e34e6a2af05da6c4d70) [signature tests](https://github.com/jsonata-js/jsonata/commit/ff36f0bd0f1aa4307662ffcc9f68abbba2f20915) when I stumbled upon something that looks like an unexpwected behaviour to me:

https://github.com/jsonata-js/jsonata/blob/597e5ee6ada3e13eaa4880f00468dcc1cba21142/src/signature.js#L230-L254

specifically, if there's an empty `match` for a `param` and `param` is not a context/contextRegex then the function would just silently validate current `arg` for the `param`. 

This seem to be plainly wrong, and it would cause stuff like this:

```
λ($arg1, $arg2)<s?n?:o>{{"$arg1": $arg1, "$arg2": $arg2}}(2)
```
produces
```json
{
  "$arg1": 2
}
```

instead of
```json
{
  "$arg2": 2
}
```

https://try.jsonata.org/KS7syiOV_

And 

```
λ($arg1, $arg2)<s?n?:o>{{"$arg1": $arg1, "$arg2": $arg2}}(1,2)
```
produces error (expected) with unexpected message: `Argument 2 of function undefined does not match function signature` — one would expect that it is argument 1 that does not match the signature.

PR #777, opened 2026-03-30
Attempt to properly implement `?` and `+` signature semantics

mikhail-barg:
This PR attempts to address #776

I'm referring to [the docs on `+` and `?`](https://github.com/mikhail-barg/jsonata/blob/3d017f42226c14e4520ee8dcf7fd68a7841a0bd5/docs/programming.md?plain=1#L257-L265) and my interpretation of these lines is as follows: 
* if an argument with `?` does not match anything, it gets no value => gets `undefined`;
* if an argument with `+` matches multiple values, it gets an arry of these values as a value (because there's no other way to pass multiple values into a single argument), so:
  * `n+` matching `1.0` would result in a literal `1.0`
  * `n+` matching `1.0, 2.0, 3.0` woul result in array `[1.0, 2.0, 3.0]`

As you may see, I've only changed recently added 035 and 040 tests to match the semantics above, so I believe this change is not a major breaking change. It's more like covering previously vague topic. I've also added a number of tests to display the behaviour more explicitly.

@andrew-coleman what do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "scripts": {
     "pretest": "npm run lint",
     "mocha": "nyc ./node_modules/mocha/bin/_mocha -- \"test/**/*.js\"",
+    "mocha2": "nyc ./node_modules/mocha/bin/_mocha -- \"test/signature2-tests.js\"",
     "test": "npm run mocha",
     "posttest": "npm run check-coverage && npm run browserify && npm run minify && npm run build-es5",
     "build-es5": "npm run mkdir-dist && npm run regenerator && npm run browserify-es5 && npm run minify-es5",
--- a/src/signature.js
+++ b/src/signature.js
@@ -248,11 +248,20 @@ const signature = (() => {
                                     };
                                 }
                             } else {
-                                validatedArgs.push(arg);
-                                argIndex++;
+                                //validatedArgs.push(arg);
+                                //argIndex++;
+                                validatedArgs.push(undefined);
                             }
+                        } else if (match.length > 1 && param.type !== 'a') { // may have matched multiple args (if the regex ends with a '+')
+                            //the signature should have been like 'n+' - we still pack multiple args into array
+                            var argsAsArray = [];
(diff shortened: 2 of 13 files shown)


Turings98apprentice:
While I would agree that the docs do not seem to match the current behavior, **I do believe this is a breaking change**, because it changes how function arguments are set. JSONata scripts written before this change would have no guarantee of working the same after this change.

I think this is a step in the right direction for function signatures, but I also think we need to examine this issue a little deeper before making v3.0. For example, there are still other [undocumented (oops)] [issues](https://try.jsonata.org/99bbwUgg0) with how function signatures work that we should make an effort to address/correct in the same version jump (imo).

mikhail-barg:
@Turings98apprentice Thanks for pointing to another unexpected signature cases. I've added cases 045-047 based on your examples, and while 045 is passing with my changes applied, 046 and 047 are not. 

I see where the problems are coming from in these cases, and I believe it's impossible to slove those with the current regex-based signature validation approach.

Since you've mentioned approaching v3.0, I think it would be possible to re-think the signature engine as a whole. So I made an mock/MVP with a custom state automata. Please see [signature2.js](https://github.com/mikhail-barg/jsonata/blob/c127bc01b9f2856432720b5766de198216a4b0ce/src/signature2.js) and [signature2-test.js](https://github.com/mikhail-barg/jsonata/blob/c127bc01b9f2856432720b5766de198216a4b0ce/test/signature2-tests.js) for the idea of the behaviour becoming possible. 

This automata approach is based on [this paper](https://swtch.com/~rsc/regexp/regexp1.html). I've had an implementation of it once before and I believe it gives quite a lot of flexibility. So what do you think?

Note: this is totally a WIP/proposal, so there are some features laking:
* no support for `()` groups
* no support for array/function `<>` arguments
but these should be straightforward to add.

I've also haven't followrd the linter guidelines yet (and I'm totally not fluent in JS at all), but I'll fix those in case this proposal gets motion.

@andrew-coleman FYI

PR #790, opened 2026-05-11
Fix $formatNumber infinite loop on zero/negative values with exponent (#785)

### LINKED ISSUE #785: $formatNumber infinite loop with zero/negative values and exponent patterns
## Summary

`$formatNumber(0, "0e0")` and `$formatNumber(-42, "0e0")` cause an infinite loop in jsonata v2.1.0. Positive numbers with the same pattern work correctly (e.g., `$formatNumber(42, "0e0")` returns `"4e1"`).

## Version

jsonata v2.1.0

## Steps to Reproduce

```javascript
const jsonata = require('jsonata');

// This works correctly:
const r1 = await jsonata('$formatNumber(42, "0e0")').evaluate({});
console.log(r1); // "4e1"

// This hangs forever (infinite loop):
const r2 = await jsonata('$formatNumber(0, "0e0")').evaluate({});

// This also hangs forever:
const r3 = await jsonata('$formatNumber(-42, "0e0")').evaluate({});
```

## Expected Behavior

- `$formatNumber(0, "0e0")` should return `"0e0"`
- `$formatNumber(-42, "0e0")` should return `"-4e1"`

These are valid XPath/XQuery `format-number` picture strings with exponent sub-patterns. The exponent format should work regardless of the sign or zero value of the input number.

## Actual Behavior

The function enters an infinite loop and never returns. The process must be killed externally.

## Analysis

The issue appears to be in the formatting loop that handles the exponent normalization for zero and negative values. When the mantissa is zero or negative, the loop that adjusts the exponent never terminates.

Note: Uppercase `E` (e.g., `"0E0"`) is not recognized as an exponent separator by the reference implementation — it is treated as a passive character, which correctly triggers D3086. Only lowercase `e` triggers the exponent path.

## Workaround

Avoid using exponent patterns (`e`) with values that may be zero or negative. Pre-check the value before formatting:

```javascript
$number > 0 ? $formatNumber($number, "0e0") : $formatNumber($abs($number), "0e0")
```

(This workaround still fails for zero.)

SAY-5:
Closes #785.

`$formatNumber(0, "0e0")` and `$formatNumber(-42, "0e0")` hang forever. The mantissa-adjustment loops in `formatNumber` (src/functions.js) compare raw signed mantissa values against the magnitude thresholds:

```js
while (mantissa < minMantissa) { mantissa *= 10; exponent -= 1; }
while (mantissa > maxMantissa) { mantissa /= 10; exponent += 1; }
```

For zero, `0 * 10 === 0` so `mantissa < 1` stays true and `exponent` underflows toward `-Infinity` without yielding to the event loop. For any negative input, `mantissa < 1` is trivially true (negatives are always less than the positive minimum).

Per XPath F&O 4.7.3 Bullet 5 the loops are meant to normalise the **magnitude** of the mantissa and the sign is already handled separately via the second sub-picture prefix. Switch the comparisons to `Math.abs(mantissa)` and short-circuit when `mantissa === 0`.

Regression test `test/test-suite/groups/function-formatNumber/issue785.json` covers zero, negative, positive, and zero-with-fractional cases — all hang on `master` (verified by interrupting an `npx mocha --grep issue785` run that ran for tens of seconds without producing output) and pass after the fix.

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -1074,13 +1074,21 @@ const functions = (() => {
             var minMantissa = Math.pow(10, pic.scalingFactor - 1);
             mantissa = adjustedNumber;
             exponent = 0;
-            while (mantissa < minMantissa) {
-                mantissa *= 10;
-                exponent -= 1;
-            }
-            while (mantissa > maxMantissa) {
-                mantissa /= 10;
-                exponent += 1;
+            // Compare magnitudes; previously `mantissa < minMantissa` was
+            // true for every negative mantissa and `0 * 10 === 0` never
+            // crossed the threshold, producing an infinite loop on zero
+            // and negative inputs (#785). For zero the desired exponent is
+            // simply zero — the XPath F&O Bullet 5 says "if N is zero, set
+            // M to zero and E to zero".
+            if (mantissa !== 0) {
+                while (Math.abs(mantissa) < minMantissa) {
+                    mantissa *= 10;
+                    exponent -= 1;
+                }
+                while (Math.abs(mantissa) > maxMantissa) {
+                    mantissa /= 10;
+                    exponent += 1;
+                }
             }
         }
         // bullet 6:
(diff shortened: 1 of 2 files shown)


mattbaileyuk (approved):
Thanks for getting to the bottom of this and contributing a fix 🙂

PR #808, opened 2026-07-01
Fix some incorrect examples

joshuaeastep74:
Two of the examples do not yield the results noted.  I fixed the example so that the result is correct.

### DIFF
--- a/website/versioned_docs/version-2.1.0/programming.md
+++ b/website/versioned_docs/version-2.1.0/programming.md
@@ -87,7 +87,7 @@ __Examples__
 
 <div class="jsonata-ex">
   <div>Account.Order.Product.{
-    `Product Name`: $.'Product Name',
+    "Product Name": $.'Product Name',
     `Category`: $.Category ?: "Uncategorized"
 }</div>
   <div>[
@@ -337,7 +337,7 @@ Consider the following example:
 Account.(
   $AccName := function() { $.'Account Name' };
 
-  Order[OrderID = 'order104'].Product.{
+  Order[OrderID = 'order104'].Product{
     'Account': $AccName(),
     'SKU-' & $string(ProductID): $.'Product Name'
   }


website/versioned_docs/version-2.1.0/programming.md
```diff
@@ -87,7 +87,7 @@ __Examples__
 
 <div class="jsonata-ex">
   <div>Account.Order.Product.{
-    `Product Name`: $.'Product Name',
+    "Product Name": $.'Product Name',
```
mattbaileyuk, line 90: I don't think this change is needed, the backticks are fine (I get the same result in try.jsonata.org with both options)

website/versioned_docs/version-2.1.0/programming.md
```diff
@@ -337,7 +337,7 @@ Consider the following example:
 Account.(
   $AccName := function() { $.'Account Name' };
 
-  Order[OrderID = 'order104'].Product.{
+  Order[OrderID = 'order104'].Product{
```
mattbaileyuk, line 340: Agree with this one that the `.` is errant (or the result is wrong, given it's just an example of context)

mattbaileyuk (changes requested):
I don't think the backtick one needs changing, so that could be removed. The dot is wrong, though.

The file here is the generated output, and given it's wrong I'm happy to change the past, but the change also needs to go into https://github.com/jsonata-js/jsonata/blob/master/docs/programming.md so it doesn't get lost in the next version.

website/versioned_docs/version-2.1.0/programming.md
joshuaeastep74: For me, the backticks output the field value (such as Bowler Hat) as opposed to what the result shows (which is Product Name) in try.jsonata.org. Category should also be surrounded with double quotes.  The original JSONata outputs this for me:

[
  {
    "Bowler Hat": "Bowler Hat"
  },
  {
    "Trilby hat": "Trilby hat"
  },
  {
    "Bowler Hat": "Bowler Hat"
  },
  {
    "Cloak": "Cloak"
  }
]

The following modification outputs the correct result:

Account.Order.Product.{
    "Product Name": $.'Product Name',
    "Category": $.Category ?: "Uncategorized"
}

mattbaileyuk: Ah, yes, you're right! That makes more sense now I'm more woken up 🙂 So line 91 also needs updating to replace the backticks with double quotes.

